PR #21, opened 2024-07-02
Fallback to vim.ui.select if Telescope is not found

julienvincent:
Currently the `set_selected_env` API only supports telescope. This changes it to fallback to using native `vim.ui.select`, which can already be mapped to other pickers by using something like [dressing](https://github.com/stevearc/dressing.nvim)


gorillamoe:
Thanks for this great addition! 👌🏾

PR #22, opened 2024-07-02
Add a warning that this only works on 0.10

julienvincent:
It took me a while to realise that the reason it was failing was because I was on 0.9.5. Having this on the front page would be useful.


gorillamoe:
Thanks for making the readme easier to understand! 💪🏾

Issue #41, opened 2024-07-10
Kulala blocks/freezes neovim waiting for http response


coltorl:
Hello, I just started using the plugin and I think it's really neat! When I make a simple GET request with a debugger running with a breakpoint on the entrance to the GET function, the whole neovim client freezes. I believe it is because of a blocking cmd run here: https://github.com/mistweaverco/kulala.nvim/blob/939025ce5357d1d534a01faf454e012d87874f59/lua/kulala/cmd/init.lua#L9

The command blocks the neovim client until the response is returned, however because I have dap running in the same frozen client, I can't unbreak out of the breakpoint so it just stays frozen.

I fixed this locally and will open a PR for you to review/improve upon.

Thanks again!

gorillamoe:
Awesome improvement! 💪🏾

PR #42, opened 2024-07-10
make cmd non-blocking

coltorl:
Addresses #41


yujinyuz:
I also tested this and seems like a good solution.

But the only problem so far is that if the API responds too long, it immediately shows the results of an old request

gorillamoe (approved):

PR #85, opened 2024-07-29
docs: fix type annotations syntax to follow LuaCATS

TheLeoP:
https://luals.github.io/wiki/annotations/#return


gorillamoe (approved):
Thanks for making these improvements <3

PR #86, opened 2024-07-30
fix(fs): avoid infinite loop searching parent dir

### LINKED ISSUE #83: Bug: [windows] while using oil.nvim `find_file_in_parent_dirs` creates an infiinte looop
Opening Neovim via `nvim .` (current working directory) while using both `kulala.nvim` and `oil.nvim` results in an infinite loop.

This happens because the function is checking `dir ~= '/' and dir:match('[A-Z]:\\') == nil`, but `dir` never gets the expected value (the values of `dir` in each iteration are showed bellow)

```
function#while dir: "oil:///C/Users/pcx/AppData/Local"
function#while dir: "oil:///C/Users/pcx/AppData"
function#while dir: "oil:///C/Users/pcx"
function#while dir: "oil:///C/Users"
function#while dir: "oil:///C"
function#while dir: "oil"
function#while dir: "."
function#while dir: "."
function#while dir: "."
...
```

Possible solutions(non exclusive):
- Use `vim.fs.find` (with a hard limit of upwards directories) instead of a while loop in `find_file_in_parent_dirs`
- Do not search for `.env` files always on startup. Do it instead on an filetype autocmd for `http` instead

I'll be happy to open a PR with any of these (or any other) solution :D

TheLeoP:
Fixes #83

### DIFF
--- a/lua/kulala/utils/fs.lua
+++ b/lua/kulala/utils/fs.lua
@@ -5,20 +5,7 @@ local M = {}
 -- @return string|nil
 -- @usage local p = fs.find_file_in_parent_dirs('Makefile')
 M.find_file_in_parent_dirs = function(filename)
-  local dir = vim.fn.expand('%:p:h')
-  -- make sure we don't go into an infinite loop
-  -- if the file is in the root directory of windows or unix
-  -- we should stop at the root directory
-  -- for linux, the root directory is '/'
-  -- for windows, the root directory is '[SOME_LETTER]:\'
-  while dir ~= '/' and dir:match('[A-Z]:\\') == nil do
-    local parent = dir .. '/' .. filename
-    if vim.fn.filereadable(parent) == 1 then
-      return parent
-    end
-    dir = vim.fn.fnamemodify(dir, ':h')
-  end
-  return nil
+  return vim.fs.find({filename}, {upward=true, limit=1})[1]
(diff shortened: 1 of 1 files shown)


gorillamoe:
Thanks for making Kulala a better tool for everyone.

I'm not sure if this is the right approach to deal with it.. If I understand it correctly it will just look up one dir, right?

If so, I wont have that PR getting merged, as is, but will propose a different solution to the problem, because the spec clearly says, that it will traverse the whole directory tree upwards, until it finds a file.

gorillamoe:
I have to correct myself; it's not the spec that says it, but some other clients do it this way e.g. [visual studio](https://learn.microsoft.com/en-us/aspnet/core/test/http-files?view=aspnetcore-8.0)

> The environment file doesn't have to be in the project folder. Visual Studio looks for an environment file in the folder where the .http file exists. If it's not in that folder, Visual Studio looks through the parent directories to find it. When a file named http-client.env.json is found, the search ends. The file found nearest to the .http file is used.

TheLeoP:
> If I understand it correctly it will just look up one dir, right?

No. It will always stop after finding the first file (hence the `limit =1`). It just handles better not getting stuck in an infinite loop

gorillamoe (approved):

PR #110, opened 2024-08-09
Allow spaces in query part from request line

Grueslayer:
The request line was spliitted by space, so it was not possible to add spaces to the query values.
Now the parts Method, Url, Version are explicitly splitted.


gorillamoe:
Shit, we have a regression. My first thought was that it has something to do with this PR, but it's present in main :see_no_evil: 

Seems fine to me.

gorillamoe (approved):

Issue #121, opened 2024-08-11
Headers with values containing a colon are not parsed correctly


felixscherz:
Hi, thank you for working on this plugin!

I was playing around with the authorization headers and found that basic auth would not work for me. This is the request I tried:

```
GET http://httpbin.org/basic-auth/foo/bar HTTP/2.1
Authorization: Basic foo:bar
```
Which using `curl -v http://httpbin.org/basic-auth/foo/bar -u foo:bar` should respond with

```json
{
  "authenticated": true,
  "user": "foo"
}
```
I investigated this a bit and found that the headers were not being parsed correctly if the value contained a colon (`:`).
Slightly modifying the regular expression used for parsing the headers fixed the issue for me, I'll submit a pull request in a second :)

gorillamoe:
You're a hero! :heart:! What a great catch! Thanks for your PR! It's already merged into main and will be released soon!

PR #122, opened 2024-08-11
fix: split headers into key value at first colon

felixscherz:
Hi, this is in regards to #121.
With this change, headers are split into key, value pairs at the first colon.

### DIFF
--- a/lua/kulala/parser/init.lua
+++ b/lua/kulala/parser/init.lua
@@ -185,7 +185,7 @@ M.get_document = function()
             request.body = request.body .. line .. "\r\n"
           end
         end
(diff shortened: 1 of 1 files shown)


gorillamoe:
What a great catch :heart_eyes: !!!

gorillamoe (approved):

estacioneto:
Hi,

First of all, nice catch indeed, @felixscherz!

`kulala.inspect()` still does not show the header so running `kulala.run()` from the computed buffer doesn't work. I'll investigate it to see if I can find where it's failing.

FYI @gorillamoe

PR #134, opened 2024-08-14
feat: add option to configure split direction

shaeinst:


shaeinst (commented):
opps we already have CONFIG, my bad.

gorillamoe (approved):

PR #142, opened 2024-08-15
Vscode rest client env support

Grueslayer:
### vscode_rest_client_environmentvars

If enabled, Kulala searches for `.vscode/settings.json` or `*.code-workspace` files in the current directory and its parents to read the `rest-client.environmentVariables` definitions (`$shared` will be treated as `_base`).

If `http-client.env.json` is also present, it will be merged (and overwrites variables from VSCode).

Possible values:

- `true`
- `false`

Default: `false`

Example:

```lua
require("kulala").setup({
  vscode_rest_client_environmentvars = true,
})
```

### DIFF
--- a/docs/docs/getting-started/setup-options.md
+++ b/docs/docs/getting-started/setup-options.md
@@ -67,7 +67,9 @@ require("kulala").setup({
     "}",
   },
   -- enable winbar
-  winbar = false;
+  winbar = false,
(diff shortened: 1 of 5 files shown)


gorillamoe (approved):

gorillamoe:
Thanks for this awesome feature! :heart:

PR #151, opened 2024-08-17
Update lazy.nvim instruction in README.md

Elnee:
First of all, thank you for such an amazing project! I want to contribute to the README.md file.  
It's usual to have a listing of just a Lua table item as an install snippet. Because many people use a rather structural approach to the configuration of `lazy.nvim`. So, the `require` is already called by them in a separate file.

Also, I removed the `config` because the [official documentation](https://lazy.folke.io/spec#spec-setup) of `lazy.nvim` suggests using `opts` instead of it. It will do the same. It will just call `require('kulala').setup({})` but with less code.

I've tested the change and it works.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,16 +38,7 @@ Via [lazy.nvim](https://github.com/folke/lazy.nvim):
 ### Configuration
 
 ```lua
-require('lazy').setup({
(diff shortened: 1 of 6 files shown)


gorillamoe:
Thanks for the great addition!

PR #153, opened 2024-08-18
Fix url parsing bug

zruhao:
Fixed the issue of Nil pointers in certain URL parsing, such as http://hostname/xxx?a=1&b


gorillamoe:
Thanks for the awesome bugfix!

gorillamoe (approved):

PR #154, opened 2024-08-18
Enable pre- and post-request scripting support.

gorillamoe:
Enable pre- and post-request scripting support.

## Example inline

```http
< {%
// some inline javascript pre-request script
%}
GET https://httpbin.org/get HTTP/1.1

> {%
// some inline javascript post-request script
%}
```

## Example file

```http
< ./pre-process.js
GET https://httpbin.org/get HTTP/1.1

> ./post-process.js
```

This matches the [this](https://www.jetbrains.com/help/idea/exploring-http-syntax.html) behaviour.


zruhao (approved):

gorillamoe:
> Is there support for using libraries in scripts? Like `moment`.
> 
> In postman, I can use it like this.
> 
> ```js
> const randomDate = require('moment')().format('YYYY-MM-DD HH:mm:ss')
> ```

If you have that package 📦 installed globally 🌎 or local to that `.http` in `node_modules` folder, then yes.

gorillamoe:
> Is there support for using libraries in scripts? Like `moment`.
> 
> In postman, I can use it like this.
> 
> ```js
> const randomDate = require('moment')().format('YYYY-MM-DD HH:mm:ss')
> ```

Should work like that: https://kulala.mwco.app/docs/scripts/overview#using-node-modules

Issue #172, opened 2024-08-21
[bug]: loop or previous error loading module 'kulala'


iamxiaojianzheng:
```http
< {%
request.variables.set('TITLE', '标题')
%}

POST https://jsonplaceholder.typicode.com/posts
Content-Type: application/json; charset=utf-8

{"title":"{{TITLE}}","body":"中文内容","userId":1}
```

exception stack

```
E5108: Error executing lua [string ":lua"]:1: loop or previous error loading module 'kulala'
stack traceback:
	[C]: in function 'require'
	[string ":lua"]:1: in main chunk

Failed to run `config` for kulala.nvim

...Local/nvim-data/lazy/kulala.nvim/lua/kulala/utils/fs.lua:179: attempt to concatenate a nil value

# stacktrace:
  - kulala.nvim\lua\kulala\utils\fs.lua:179 _in_ **get_plugin_root_dir**
  - kulala.nvim\lua\kulala\utils\fs.lua:101 _in_ **get_scripts_dir**
  - kulala.nvim\lua\kulala\scripts\javascript.lua:5
  - kulala.nvim\lua\kulala\scripts\init.lua:1
  - kulala.nvim\lua\kulala\parser\init.lua:12
  - kulala.nvim\lua\kulala\ui\init.lua:6
  - kulala.nvim\lua\kulala\init.lua:1
  - ~/AppData/Local/nvim/lua/plugins/kulala.lua:10 _in_ **config**
```

PR #173, opened 2024-08-21
fix(fs): Compatible with "/" and "\\"

iamxiaojianzheng:


iamxiaojianzheng:
https://github.com/mistweaverco/kulala.nvim/issues/172

PR #183, opened 2024-08-23
feat(from_curl): import HTTP spec from curl command line

vyskocilm:
Curl is a standard way of specifying the API requests. It is a common format in scripts, API docs and browsers supports Copy as cURL in the network tools as well.

This commit adds

1. shlex-lua source code into project
2. code parsing the curl command line and generating the Request table
3. code appending the Request into current buffer

Author of https://github.com/BodneyC/shlex-lua states his code is a drop-in one and hasn't been updated since then. I assume it's fine to drop it into kulala code. Tested on curl line generated by kulala itself. Parser omits unknown curl options, do not support long variants (yet) and code writing spec back to neovim buffer is probably too simple. Note this is my first coding in neovim/lua, so I'd be happy for any advice improving the code.

image


lua/kulala/init.lua
```diff
@@ -10,83 +10,87 @@ local ScriptsUtils = require("kulala.scripts.utils")
 local M = {}
 
 M.setup = function(config)
-  CONFIG.setup(config)
+	CONFIG.setup(config)
```
vyskocilm, line 13: Haven't realized I have an automatic formatting enabled. Will try to disable it.

gorillamoe:
Really, really nice addition 👌🏾👍🏾.

I'm in the middle of moving, so I don't have much time ATM, but will try to have a look this evening.

Thanks 🙏🏾 for helping to make Kulala outstanding 🥳

vyskocilm:
Opened a new one https://github.com/mistweaverco/kulala.nvim/pull/186 - with a clean formatting. I added stylua somewhere in the past and haven't realized this is not the default.

PR #184, opened 2024-08-23
fix: multipart/form-data

philopence:
Hi, I simply fixed the option of `multipart/form-data`

Now it works with single field

(`multipart/form-date` requires the `-F/--form` option)

Also, the command to submit multiple fields is: `curl -F name=John -F shoesize=11 https://example.com/`

But I don't know how to implement it :(


gorillamoe:
Superseeded by #230

PR #186, opened 2024-08-24
feat(from_curl): import HTTP spec from curl command line

vyskocilm:
Curl is a standard way of specifying the API requests. It is a common format in scripts, API docs and browsers supports Copy as cURL in the network tools as well.

This commit adds

1. shlex-lua source code into project
2. code parsing the curl command line and generating the Request table
3. code appending the Request into current buffer


lua/kulala/parser/curl.lua
```diff
@@ -0,0 +1,78 @@
...
+          res.method = "POST"
+        end
+      elseif arg == "--http1.1" then
+          res.http_version = "HTTP/1.1"
+      elseif arg == "--http2" then
+          res.http_version = "HTTP/2"
+      elseif arg == "--http3" then
+          res.http_version = "HTTP/3"
+      end
+      goto continue
```
vyskocilm, line 53: Not sure if this is safe to use in neovim Lua plugins though.

lua/kulala/ui/init.lua
```diff
@@ -118,6 +119,28 @@ local function pretty_ms(ms)
...
+    idx = idx + 1
+    lines[idx] = header .. ": " .. value
+  end
+  if spec.body ~= "" then
+    idx = idx + 1
+    lines[idx] = ""
+    -- FIXME: broken for multi-line body
+    lines[idx+1] = spec.body
+  end
+  vim.api.nvim_put(lines, 'l', false, false)
```
vyskocilm, line 141: Dunno if this is a good approach, however is easier to use than nvim_buf_set_lines, which overwrite the content of the file.

lua/kulala/parser/curl.lua
```diff
@@ -0,0 +1,78 @@
+local Shlex = require("kulala.shlex")
+
+local M = {}
+
+---Parse a curl command into a Request object
+---@param curl string The curl command line to parse
+---@return Request|nil -- Table with a parsed data or nil if parsing fails
```
vyskocilm, line 7: I get an error for underfined Request. Can I use the one from `init.lua` or better define own type here?

lua/kulala/parser/curl.lua
```diff
@@ -32,7 +32,10 @@ function M.parse(curl)
 
   for _, arg in ipairs(parts) do
     if state == State.START then
-      if arg == "-X" or arg == "--request" then
+      if arg:match("^http[s]?://") and res.url == "" then
```
gorillamoe, line 36: Wouldn't it be better to make it match `://` so we are maybe more future-proof?

lua/kulala/parser/curl.lua
gorillamoe: I think you can use the one from `init.lua`.

lua/kulala/parser/curl.lua
vyskocilm: https://github.com/mistweaverco/kulala.nvim/pull/186/commits/7f31918e127d0b85a292db9c8c3b44f3b80b3a50

lua/kulala/parser/curl.lua
vyskocilm: I am still new to Lua - I wanted to share annotations, which sounds like a complicated topic in Lua - https://github.com/LuaLS/lua-language-server/wiki/Annotations. Maybe I'll drop the annotations from the code. They're not important for feature to work.

gorillamoe:
Works with Firefox, but not (yet) with Chrome.

gorillamoe:
Thanks for this awesome contribution! :heart:

Issue #188, opened 2024-08-24
[Feature UI] Add console in winbar


iamxiaojianzheng:
```http
POST https://jsonplaceholder.typicode.com/posts
Content-Type: application/json; charset=utf-8

{"title":"中文"}

> {%
client.log(response.body.title)
%}
```

Hope to add a `Console` panel to view the print output from `pre-request` and `post-request`.

![image](https://github.com/user-attachments/assets/79e52c2d-324b-4d0a-aeca-29b3d27242e3)

Related to https://github.com/mistweaverco/kulala.nvim/issues/176

gorillamoe:
Why'd you close the PR?

PR #189, opened 2024-08-24
feat(ui): add console winbar

iamxiaojianzheng:
```http
< {%
console.log(Math.random())
%}

POST https://jsonplaceholder.typicode.com/posts
Content-Type: application/json; charset=utf-8

{"title":"{{$uuid}}"}

> {%
console.log(response.body)
%}
```

image

Related to https://github.com/mistweaverco/kulala.nvim/issues/188


gorillamoe:
I would like to have this optional though, I mean that you can opt-out of showing console messages right away.

I find it useful to have them available via `:messages`, but I think that some people might feel annoyed by them, so we can make it optional.

I'll work on that right away.

PR #190, opened 2024-08-25
feat(telescope): add telescope sorter

imiljan:
If telescope is installed, with this change it will be possible to filter the output when using `require('kulala').search()` and `require('kulala').set_selected_env()`


gorillamoe (approved):
Nice work :+1: :muscle:

PR #197, opened 2024-08-31
Feature/add console winbar

iamxiaojianzheng:
1. Add console winbar
2. Make winbar panes configurable

about this https://github.com/mistweaverco/kulala.nvim/issues/188

### DIFF
--- a/docs/docs/getting-started/configuration-options.md
+++ b/docs/docs/getting-started/configuration-options.md
@@ -13,13 +13,17 @@ Here is a full example of setting up the Kulala plugin with the available `opts`
     -- split direction
     -- possible values: "vertical", "horizontal"
     split_direction = "vertical",
+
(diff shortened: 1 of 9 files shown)


iamxiaojianzheng:
@gorillamoe Sorry, I accidentally messed up the branch, so I recreated it.

gorillamoe:
I would like to propose a renaming of "console" to "script output", because the plan is to have support for multiple scripting languages at one point and not every language has the console notation. For some it's just print or echo or something else.

Also I want to make it optional to suppress vim.print statements.

I'll rework this PR a bit, but all in all pretty good work :muscle: :+1:

Thanks for your contribution!

gorillamoe:
Thanks for your hard work :heart:

gorillamoe (approved):

PR #207, opened 2024-09-02
Feat/quick run request

iamxiaojianzheng:
It is easier to execute an http request by simply pressing Enter.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+root = true
+
+[*]
+insert_final_newline = true
+indent_style = space
+indent_size = 2
+charset = utf-8
+
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @gorillamoe @Grueslayer @darteil @TheLeoP @caenrique
+* @gorillamoe @Grueslayer @treywood
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,3 +1,4 @@
+---
 changelog:
   exclude:
     labels:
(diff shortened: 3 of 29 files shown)


gorillamoe:
Being honest I really dislike the idea of having default keybindings for stuff, even if it's somehow optional through configuration. I think that should be better left to the user and their configuration of neovim.

But I'm open for discussion, if @Grueslayer and @treywood both find this useful, we merge it.

Grueslayer:
@gorillamoe I’m on your side: let the user define its own

iamxiaojianzheng:
I think plug-ins that provide a relatively good default way to interact, while opening up the user to modify the configuration, are the right thing to do.

Users of neovim and lua may not be particularly familiar with these things. I was from the little white step by step.

There are some features and uses that not everyone knows how to use.

gorillamoe:
> Users of neovim and lua may not be particularly familiar with these things. I was from the beginner step by step.
>
> There are some features and uses that not everyone knows how to use.

Yes, but I think that this is a different topic and needs to be adressed some other way.

Mayebe we need a dedicated section for setting up keybinds in the docs.

I think that would be perfect, because you're totally right, not everyone knows this stuff and wants to mess with this stuff too.

iamxiaojianzheng:
It is also OK

Provide a best practice for users who are new to kulala. 

Let them experience the best way to use it from the very beginning.

gorillamoe:
As a contributor of some really good PRs @iamxiaojianzheng, maybe you can rework that PR into a "improve documenation" PR, if you like?

iamxiaojianzheng:
> As a contributor of some really good PRs @iamxiaojianzheng, maybe you can rework that PR into a "improve documenation" PR, if you like?

I think that's fine, and I think we can add a new "Best Practices" section to "Getting Started" to show the configuration methods we recommend to users.

PR #210, opened 2024-09-02
fix(ui/parser): remove unused variable, remove nil warnings

vyskocilm:
- fix(ui): remove unused variable
- fix(parser/ui): copy throws an error if no HTTP request is specified


gorillamoe (approved):
LGTM :+1:

Issue #209, opened 2024-09-02
[Feature] Virtual text display variable


iamxiaojianzheng:
@gorillamoe I wish add this feature. What do you think?

I've done the basic implementation locally

![image](https://github.com/user-attachments/assets/a6dbb21d-2180-4d47-a15d-dec79e9dd298)

gorillamoe:
Ye, looks very nice, but make it optional and opt-in, not opt-out, please.

iamxiaojianzheng:
> Ye, looks very nice, but make it optional and opt-in, not opt-out, please.

I added the toggle method

Grueslayer:
Would like to have an option for showing this as some “hover text” for the variable where the cursor is currently on only.

Otherwise all variables for complete file has to be calculated all the time and personally I dislike the code widening.

Also the text needs to be limited, bearer tokens can be multiple 1000 chars long.

gorillamoe:
Will start working on hover again soon. Closing this as the PR was pretty outdated.

igorlfs:
Could we have that with the in-process LSP + inlay hints?

(apologies for the necrobump)

gorillamoe:
To be honest I really hate this "feature", but since I've seen this becoming popular lately and some of my coworkers having this enabled by default for basically everything and I have to fight all LSPs to disable this now default way, I'll implement it as opt-in!

igorlfs:
> I'll implement it as opt-in

Thanks, you're the GOAT!

> To be honest I really hate this "feature"

I also think inlay hints are incredibly noisy, if left on all the time. In my config, they are disabled but [I have a mapping](https://github.com/igorlfs/dotfiles/blob/e90cfceea77b49145b39d0278497a7226df5efa1/nvim/.config/nvim/lua/igorlfs/autocmds.lua#L149-L151) to quickly toggle them when I need to.

gorillamoe:
> I also think inlay hints are incredibly noisy, if left on all the time. In my config, they are disabled but [I have a mapping](https://github.com/igorlfs/dotfiles/blob/e90cfceea77b49145b39d0278497a7226df5efa1/nvim/.config/nvim/lua/igorlfs/autocmds.lua#L149-L151) to quickly toggle them when I need to.
> 


TBH, it never occurred to me that I could just toggle them .. I think I might steal this from you.

gorillamoe:
[...]

PR #211, opened 2024-09-02
feat(inlay): show virtual variable text

iamxiaojianzheng:
about this https://github.com/mistweaverco/kulala.nvim/issues/209


gorillamoe:
This assumes people using TS. I would like to have this optional. I can look into some refactoring for that.

lua/kulala/inlay/init.lua
```diff
@@ -44,4 +51,66 @@ M.show = function(t, linenr)
...
+
+M.toggle_virtual_variable = function()
+  M.show_virtual_variable_text = not M.show_virtual_variable_text
+
+  if M.show_virtual_variable_text then
+    add_virtual_variable_text()
+    vim.api.nvim_create_augroup(VV_GROUP_NAME, { clear = true })
+    vim.api.nvim_create_autocmd({ "BufEnter", "TextChanged", "TextChangedI" }, {
+      group = VV_GROUP_NAME,
+      pattern = "*.http",
```
gorillamoe, line 108: I think 🤔 we might also include *.rest here

gorillamoe (changes requested):

iamxiaojianzheng:
> This assumes people using TS. I would like to have this optional. I can look into some refactoring for that.

Look forward to your optimization

lua/kulala/inlay/init.lua
iamxiaojianzheng: has done

PR #212, opened 2024-09-02
feat(ci): test support via plenary test harness

vyskocilm:
* add a first test for curl command line parsing.
 * add a GHA workflow file

Inspired by https://github.com/nvim-neotest/neotest/blob/master/.github/workflows/workflow.yaml


.github/workflows/tests.yml
```diff
@@ -0,0 +1,43 @@
...
+
+  tests:
+    name: tests
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-22.04
+            rev: v0.10.0/nvim-linux64.tar.gz
```
vyskocilm, line 16: `neotest` does testing on a nightly branch. I don't think this is so interesting, so I removed the nightly branch. The way how `neotest` did this means the cache key is invalidated every day, making cache action less useful.

lua/kulala/parser/curl.lua
```diff
@@ -89,7 +92,7 @@ function M.parse(curl)
         res.headers["user-agent"] = arg
       elseif state == State.Header then
         local header, value = Stringutils.cut(arg, ":")
-        res.headers[Stringutils.remove_extra_space(header)] = Stringutils.remove_extra_space(value)
+        res.headers[Stringutils.remove_extra_space(header):lower()] = Stringutils.remove_extra_space(value)
```
gorillamoe, line 95: I don't think we can do that, because that would be a regression to stuff that @Grueslayer fixed, right?

gorillamoe (changes requested):

lua/kulala/parser/curl.lua
gorillamoe: Should be fixed in 6760d2c

gorillamoe (dismissed):

gorillamoe (approved):

gorillamoe (dismissed):

gorillamoe (dismissed):

gorillamoe (approved):

gorillamoe (dismissed):

gorillamoe (dismissed):

gorillamoe (dismissed):

treywood (approved):

gorillamoe (approved):

gorillamoe (dismissed):

gorillamoe (dismissed):

gorillamoe (approved):

PR #216, opened 2024-09-06
Environment Scope

Grueslayer:
Pushes request responses and environment variables to vim buffer variables, so each file can have different values for same requests variable. That is the same behaviour like vscode-rest-client: scope of variables and requests is per file.

You can switch back to "global" scope by config.

DEV: Keep in mind to use `DB.get()` instead of `DB.data`


gorillamoe:
Does this break `.replay()`? Have to check 👀

gorillamoe (dismissed):

gorillamoe:
I also added tests for the DB change I introduced and made sure it works as expected:
https://github.com/mistweaverco/kulala.nvim/actions/runs/10751816491/job/29819466848

PR #220, opened 2024-09-08
feature: show variable values on hover

gorillamoe:


iamxiaojianzheng:
I don't think the current pop-up display is ideal for interaction.

When the cursor moves to `Host`

image image  

Could replace it like this ?

image

iamxiaojianzheng:
Is there any way to edit this branch together?

gorillamoe:
> I don't think the current pop-up display is ideal for interaction.
> 
> When the cursor moves to `Host`
> 
> image image
> 
> Could replace it like this ?
> 
> image

Yes, true, I'm still working on this. Needs some more love I guess :+1:

gorillamoe:
@Grueslayer, can you check if this works for you? I'm pretty satisfied with the current implemenation.

```lua
return {
  "mistweaverco/kulala.nvim",
  opts = {
    show_variable_info_text = "float",
  },
}

```

gorillamoe:
Env parsing ist not working correctly as it seems, but your 2 cents on the float would be great :)

gorillamoe:
image

PR #225, opened 2024-09-09
[not urgent] enable treesitter parsing by default

treywood:
I've been thinking about how to start enabling tree-sitter parsing by default, and I think this might be a good way to go about it - the `treesitter` option is set to true by default if the user has an http parser installed, and documentation encourages the user to try setting this option to false if they encounter an issue.

No rush to merge this, but curious about thoughts on the approach

### DIFF
--- a/docs/docs/getting-started/configuration-options.md
+++ b/docs/docs/getting-started/configuration-options.md
@@ -93,6 +93,11 @@ Here is a full example of setting up the Kulala plugin with the available `opts`
     -- set scope for environment and request variables
     -- possible values: b = buffer, g = global
     environment_scope = "b",
+
+    -- by default, requests are parsed using treesitter if an http parser is available.
+    -- if you encounter tree-sitter related issues, you can disable this feature
+    -- by setting this option to false (and then open an issue on GitHub!)
+    treesitter = true,
   },
 }
 ```
--- a/lua/kulala/config/init.lua
+++ b/lua/kulala/config/init.lua
@@ -1,6 +1,18 @@
 local FS = require("kulala.utils.fs")
 local M = {}
 
+-- enable treesitter parsing by default if http parser is available
+local treesitter_default = false
+local ok, parsers = pcall(require, "nvim-treesitter.parsers")
+if ok then
+  for _, parser in pairs(parsers.available_parsers()) do
+    if parser == "http" then
+      treesitter_default = true
+      break
+    end
+  end
+end
+
 M.defaults = {
   -- split direction
   -- possible values: "vertical", "horizontal"
@@ -64,7 +76,7 @@ M.defaults = {
   -- enable reading vscode rest client environment variables
   vscode_rest_client_environmentvars = false,
   -- parse requests with tree-sitter
-  treesitter = false,
+  treesitter = treesitter_default,
   -- disable the vim.print output of the scripts
   -- they will be still written to disk, but not printed immediately
   disable_script_print_output = false,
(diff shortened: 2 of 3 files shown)


gorillamoe:
I want to know @Grueslayer 's take on that.

I'm glad that we don't rely on ts at all for parsing, so we can implement features completely independent on our own, but maybe I'm just an old grumpy man :see_no_evil:

Grueslayer:
As long as @gorillamoe does not take care of a correct and up-to-date treesitter parser matching the Kulala syntax, I think parsing by TS should be opt-in only.

Let's have a look at this running (stripped down but looks identically with all request variables used) example:

kulala

The URLs are not correct parsed by treesitter: Kulala will (like vscode-rest-client ans others) urlencode query parameters so we can use spaces, ampersands in it's native way. TS http parser at least has a problem with the single quotes here. Maybe in that case it's cosmetic only.

gorillamoe:
To be honest, when Kulala started in v1, we had full reliance on TS and that's when so many people complained about stuff not working as expected, because TS was error'ing even for basic stuff.

That days might be gone and stuff might be fixed, but the biggest downside I see is that when we want to add additional stuff that e.g. jetbrains supports, but isn't part of the http grammar, we can't just implement and ship it, because we have to wait for our PR to get approved. And worst thing would be that the PR gets rejected, which would be totally fair, because http grammar is part of the rest.nvim project, so they can decide what goes in.

treywood:
my only real opinion (as a very new user and contributor to this project) - and the reason why I felt like utilizing treesitter here would be a worthwhile effort - is that rolling your own parsing logic with only the tools available in vimscript/lua seems unpleasant and tree-sitter makes adding/changing parser logic a lot more straightfoward, especially with a lot of the groundwork already done.

in my experience contributing to the http parser, @boltlessengineer seems very amenable to aligning with the jetbrains spec and has been quick to get my changes merged so I don't really think there would be too much friction in the future if more changes were needed to add important features to this project.

gorillamoe:
Okay, let's try this then, but first we need to make sure gql in http in  treesitter. I use it on a daily basis and this needs to work for me reliable.

When working on https://github.com/rest-nvim/rest.nvim/pull/452, I noticed that at least one of my queries failed in rest.nvim. I haven't checked if it's due to treesitter or just related to rest.nvim.

Queries are from work stuff, so I can't exactly share it as it is, but probably supply a repro example.

I think it's due to variable replacement.

I also noticed rest and kulala are quite different 🤔 which is a good thing on its own, but I struggled to get my login query to work.

Maybe I did something wrong, but named requests with response.body.$.token wasn't working for me.

Isn't this (yet) supported by rest @NTBBloodbath @boltlessengineer?

I'm referring to this here: https://kulala.mwco.app/docs/usage/request-variables

gorillamoe:
Maybe @Grueslayer can also did some quick tests on his stuff with treesitter enabled.

boltlessengineer:
> named requests with response.body.$.token wasn't working for me.

`rest.nvim` won’t support them because thats not a thing from intellij’s spec. Maybe it will be an opt-in feature in future.

Also, you won’t get answer by asking to ntb. He is completely out of rest.nvim development now.

gorillamoe:
> rest.nvim won’t support them because thats not a thing from intellij’s spec. Maybe it will be an opt-in feature in future.

Ah, thanks for the clarification. I thought that it came from intellij, but maybe it's just a thing form the vscode-rest-client by @Huachao :thinking:

PR #229, opened 2024-09-10
fix(shlex): make split function more robust

vyskocilm:
And more close to Python original. Now it supports like `curl
https://example.com` (without quotes), or ` curl 'https://example.com'`
(with a space at the start).

1. Add a test data from CPython test_shlex.py
2. Reimplement Python code generating a test cases.
3. Mark a few tests as broken and move to separate test case - probably
   not worth fixing.
4. Add a fix for curl case - set whitespace_fix to true, this is what
   Python shlex.split does.
5. Avoid `goto continue` by introducing a continue boolean flag in a
   parsing code. Resulting code is ugly, but it works.


gorillamoe (approved):
Great work, thanks! 👍🏾🙏🏾

PR #238, opened 2024-09-12
fix(from_curl): improve heuristic and recognize curl_path too

vyskocilm:


lua/kulala/parser/curl.lua
```diff
@@ -27,9 +28,9 @@ function M.parse(curl)
   curl = string.gsub(curl, "%s+", " ")
 
   local parts = Shlex.split(curl)
-  -- if string doesn't start with curl, return nil
+  -- if string doesn't start with curl or different from curl_path, return nil
   -- it could also be curl-7.68.0 or something like that
-  if string.find(parts[1], "^curl.*") == nil then
+  if string.find(parts[1], "^curl.*") == nil and parts[1] ~= Config.get().curl_path then
```
vyskocilm, line 33: I am not sure this does make a sense - maybe plugin can ignore the command name and assume this is curl :shrug:

gorillamoe: I personally find this to be a good addition. I just have one curl binary installed, but I think that some people might benefit from that.

gorillamoe (approved):

PR #241, opened 2024-09-15
Fix request variables documentation example

vlivanov1232:
Without `Content-Type` header httpbin does not parse request json and as a result `REQUEST_TWO` returns a warning that `{{REQUEST_ONE.response.body.$.json.token}}` is not found.


gorillamoe (approved):
Great 👍🏾, thanks a lot! ❤️

PR #247, opened 2024-09-17
SSL Client Certificate Auth

### LINKED ISSUE #245: [Feature Request] Allow passing certificate and key to request
I often need to pass certificates and keys when calling my APIs. Right now, I have to use curl directly with `--cert` and `--key`, but it would be awesome if Kulala supported specifying a path for both files.

Maybe something along the lines of
```http
# @cert <certificate_path>
# @key <key_path>
GET https://example.com
```
similar to how it currently handles naming and external variable parsing.

Grueslayer:
fixes #245

### DIFF
--- a/docs/docs/getting-started/configuration-options.md
+++ b/docs/docs/getting-started/configuration-options.md
@@ -618,4 +618,14 @@ Example:
 }
 ```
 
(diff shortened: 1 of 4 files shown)


gorillamoe (approved):
👌🏾❤️

Issue #128, opened 2024-08-13
toggle_view treesitter error


yujinyuz:
I'm getting this error when running `toggle_view()`

This happens when switching from Body to Header view

```lua
Error executing vim.schedule lua callback: ...table/share/nvim/runtime/lua/vim/treesitter/language.lua:107: no parser for 'pl
aintext' language, see :help treesitter-parsers
stack traceback:
        [C]: in function 'error'
        ...table/share/nvim/runtime/lua/vim/treesitter/language.lua:107: in function 'add'
        ...e/share/nvim/runtime/lua/vim/treesitter/languagetree.lua:111: in function 'new'
        .../neovim/stable/share/nvim/runtime/lua/vim/treesitter.lua:41: in function '_create_parser'
        .../neovim/stable/share/nvim/runtime/lua/vim/treesitter.lua:108: in function 'get_parser'
        ...m/stable/share/nvim/runtime/lua/vim/treesitter/_fold.lua:117: in function 'compute_folds_levels'
        ...m/stable/share/nvim/runtime/lua/vim/treesitter/_fold.lua:380: in function 'fn'
        ...m/stable/share/nvim/runtime/lua/vim/treesitter/_fold.lua:293: in function <...m/stable/share/nvim/runtime/lua/vim/
treesitter/_fold.lua:289>
```

Here's my config if it helps

```lua
    'mistweaverco/kulala.nvim',
    ft = 'http',
    opts = {
      default_view = 'headers_body',
      winbar = true,
      contenttypes = {
        ['text/plain'] = {
          formatter = function(body)
            return body
          end,
        },
      },
    },
    keys = {
      {
        '<CR>',
        function()
          require('kulala').run()
        end,
        ft = 'http',
      },
      {
        '<leader>r',
        function()
          require('kulala').run()
        end,
        ft = 'http',
      },
      {
        '<leader>v',
        function()
          require('kulala').toggle_view()
        end,
        ft = 'http',
      },
    },
```

yujinyuz:
It was a bug in my config..

yujinyuz:
My config contains

```lua
vim.opt.foldmethod = 'expr'
vim.opt.foldexpr = 'v:lua.vim.treesitter.foldexpr()'
```
Removing that resolves this error.

gorillamoe:
Can you look if this fixes your issue? https://github.com/neovim/neovim/issues/23964#issuecomment-1584411926

gorillamoe:
[...]

PR #249, opened 2024-09-29
fix(ui): treesitter foldexpr calculate not work

wenjinnn:
When setup
``` lua
vim.opt.foldmethod = 'expr'
vim.opt.foldexpr = 'v:lua.vim.treesitter.foldexpr()'
```
the fold level per lines not getting correct result, this means that `zc` `zo` will not work in result buffer. this commit fix it. 
Related #128

### DIFF
--- a/lua/kulala/ui/init.lua
+++ b/lua/kulala/ui/init.lua
@@ -108,13 +108,14 @@ vim.api.nvim_create_autocmd("WinClosed", {
 local function set_buffer_contents(contents, ft)
   if buffer_exists() then
     local buf = replace_buffer()
-    local lines = vim.split(contents, "\n")
-    vim.api.nvim_buf_set_lines(buf, 0, -1, false, lines)
+    -- setup filetype first so that treesitter foldexpr can calculate fold level per lines
     if ft ~= nil then
       vim.bo[buf].filetype = ft
     else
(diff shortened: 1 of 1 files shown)


gorillamoe:
Do we really need that replacement of buffer with this fix anymore? 😲 Would be awesome if this could be removed as well then 👌🏾👍🏾💪🏾

gorillamoe:
Seems like we still need that replace buffer :shrug:

gorillamoe (approved):

wenjinnn:
Well, I don't understand why we don't need this. Many http request return a lot of content. Using treesitter foldexpr to calculate folding can conveniently fold the content according to the syntax tree. It's useful for response content of json or xml type.
image

Imagine that your request returns thousands of lines of HTML like this, including hundreds of lines of head tag. You want to quickly view the content in the body tag. In this case, the best way is to move the cursor to the head tag and then fold the content by type `zc'. Without this fix, it will not work :).

gorillamoe:
It's already merged into main. I never said we don't need this. I was just thinking if we still need to replace the buffer everytime :)

wenjinnn:
Just saw it merged when I posting the previous comment. Yes I think we need it :).

PR #252, opened 2024-09-30
Improve scripts

gorillamoe:


gorillamoe:
Something like this is now easily possible:

`./../scripts/pre-token-gen.js`

```javascript
// https://www.jetbrains.com/help/idea/http-response-reference.html#request-properties
const url = new URL(request.url.tryGetSubstituted());
const method = request.method;
const params = new URLSearchParams(request.url.query);
const unix_timestamp = Math.floor(Date.now() / 1000);
const token_raw = client.variables.get('TOKEN_RAW');
const key1 = params.get('key1');
const computed_token = `${method}${token_raw}${key1}${unix_timestamp}`;
// Either use request.variables.set which is only valid for the current request
// or use client.global.set("COMPUTED_TOKEN", computed_token) to store the value globally
// and persist it across restarts
request.variables.set('COMPUTED_TOKEN', computed_token);
client.log({ url, method, params });

```

`advanced-scripting.http`

```http
< ./../scripts/pre-token-gen.js
POST https://httpbin.org/post?key1=value1 HTTP/1.1
Accept: application/json
Token: {{COMPUTED_TOKEN}}

{
  "foobar": "lul"
}
```

Issue #194, opened 2024-08-29
Expose request verb, body, url, query params and headers  in pre/post scripts.


hsanson:
In pre/post scripts the request seems to only have functions to set/get  variables . I have some services that use a custom HMAC authentication that requires computation of a signature that is then sent in the Authorization header.

The signature is computed by generating a canonical string that contains the request verb, url, query params, body, and a timestamp that is then encrypted with a token.  Details can be found in the [API-AUTH](https://www.rubydoc.info/gems/api-auth/2.5.1) ruby gem, that is what the service I use uses to implement authentication.

If the request object in pre/post scripts exposes this information it would be easy for me to implement the signature and add the header to the request before it is sent to the server.

gorillamoe:
Just a quick heads up. Development should start this weekend 🥵

gorillamoe:
@hsanson, with the current open PR this here is easily achievable:

Something like this is now easily possible:

`./../scripts/pre-token-gen.js`

```javascript
// https://www.jetbrains.com/help/idea/http-response-reference.html#request-properties
const url = new URL(request.url.tryGetSubstituted());
const method = request.method;
const params = new URLSearchParams(request.url.query);
const unix_timestamp = Math.floor(Date.now() / 1000);
const token_raw = client.variables.get('TOKEN_RAW');
const key1 = params.get('key1');
const computed_token = `${method}${token_raw}${key1}${unix_timestamp}`;
// Either use request.variables.set which is only valid for the current request
// or use client.global.set("COMPUTED_TOKEN", computed_token) to store the value globally
// and persist it across restarts
request.variables.set('COMPUTED_TOKEN', computed_token);
client.log({ url, method, params });

```

`advanced-scripting.http`

```http
< ./../scripts/pre-token-gen.js
POST https://httpbin.org/post?key1=value1 HTTP/1.1
Accept: application/json
Content-Type: application/json
Token: {{COMPUTED_TOKEN}}

{
  "foobar": "lul"
}
```

Will merge it today. Let me know if this solves your issue.

gorillamoe:
Docs are here: https://kulala.mwco.app/docs/scripts/request-reference

hsanson:
@gorillamoe thanks alot for implementing this feature. With this I can finally stop using Insomnia/Postman in my work flow.  Unfortunatelly I must be doing something wrong since I am unable to get this to work.

I have this http request for login that after successful login it should set a session ID and Token in the global context:

```http
[...]

PR #262, opened 2024-10-06
feat(scripts): add `request.body.getComputed()`

gorillamoe:
Returns the `string` request body as sent via curl; with variables substituted, or `undefined` if there is no body.

Very useful if you want to see the request body as it was sent to the server.

The `tryGetSubstituted` method will substitute variables with their values, but leave the rest of the body as is.

If you have a GraphQL query in the body, for example, the `getComputed` method will show the query as it was sent to the server, which is quite different from the substituted version.

As an example, if you have a request body like this:

```graphql
query getRestClient($name: String!) {
  restclient(name: $name) {
    id
    name
    editorsSupported {
      name
    }
  }
}

{
  "variables": {
    "name": "{{ENV_VAR_CLIENT_NAME}}"
  }
}
```

Then the `getComputed` method will
return the body as it was sent to the server:

```json
{"query": "query getRestClient($name: String!) { restclient(name: $name) { id name editorsSupported { name } } } ", "variables": {"variables": {"name": "kulala"}}}
```

whereas the `tryGetSubstituted` method will
return the body with variables substituted as seen in your script:

```graphql
query getRestClient($name: String!) {
  restclient(name: $name) {
    id
    name
    editorsSupported {
      name
    }
  }
}

{
  "variables": {
    "name": "kulala"
  }
}
```

The `getComputed` method is always `undefined` for binary bodies.

### DIFF
--- a/docs/docs/scripts/request-reference.md
+++ b/docs/docs/scripts/request-reference.md
@@ -25,7 +25,8 @@ client.log(request.variables.get("SOME_TOKEN"));
 
 ## request.body.getRaw
 
-Returns the request body in the raw format:
+Returns the request body `string` in
+the raw format (or `undefined` if there is no body).
 
 If the body contains variables,
 their names are displayed instead of their values.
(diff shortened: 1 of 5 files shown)


gorillamoe:
X-Ref: #194

Issue #276, opened 2024-10-08
Add content-type header for gql schema download 


gorillamoe:
@vyskocilm found out that if you're missing the content-type header in the request to download gql schema does not work.

Kulala should handle this gracefully via adding the header silently to the request if not set.

PR #277, opened 2024-10-09
fix(graphql): add a Content-Type: application/json if not specified

vyskocilm:
Fixes https://github.com/mistweaverco/kulala.nvim/issues/276


gorillamoe (approved):
👌🏾🤩

PR #279, opened 2024-10-09
fix(ui): folds not working in lazy

gorillamoe:
remove nofile, otherwise ts folds do not work in lazy.

### DIFF
--- a/lua/kulala/ui/init.lua
+++ b/lua/kulala/ui/init.lua
@@ -52,9 +52,7 @@ local replace_buffer = function()
   local old_bufnr = get_buffer()
 
   local new_bufnr = vim.api.nvim_create_buf(true, false)
-  vim.api.nvim_set_option_value("buftype", "nofile", {
-    buf = new_bufnr,
-  })
+  vim.bo[new_bufnr].swapfile = false
 
   if old_bufnr ~= nil then
     for _, win in ipairs(vim.fn.win_findbuf(old_bufnr)) do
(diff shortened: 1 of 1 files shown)


gorillamoe:
Should fix: https://github.com/mistweaverco/kulala.nvim/discussions/268

Grueslayer:
That shouldn’t be the solution. We should talk with @folke about his reason to exclude this buftypes. I don’t use LazyVim but have some more plugins with scratch buffers using buftype nofile where treesitter runs fine and I see no reason to skip folding.

gorillamoe:
Yes, folding works for me too, I don't use Lazy either, so maybe folke can elaborate on why'd he chose to disable that on nofile :thinking:

rogeriofrsouza:
Can we solve it with a user configuration on setup?

Folke is on vacation till the end of the month 😞

gorillamoe:
> @rogeriofrsouza you already confirmed that everything runs fine for you with this commit, didn’t you? So we can keep it as is (I haven’t tested for side effects). If Folke will change his repo we can do a revert (if @gorillamoe want to do it). So in the meantime everything should be good.

Yes, I will revert this ofc, if behaviour in lazy changes.
Until then it should stay as is if nobody complains about side-effects. I haven't noticed side-effects for now, but let's see.

gorillamoe:
> Yes @Grueslayer it is running fine. But I don't want to break someone else configuration out there...

Don't worry, if it breaks stuff, we'll revert it and make it opt-in.

Grueslayer:
@gorillamoe It is fixed in LazyVim, may you revert this (and maybe #286, too)?

Issue #282, opened 2024-10-10
Put scripts in a different (fs) location


willruggiano:
For those of us that have our plugins installed in a read-only filesystem.

After updating the plugin a few days ago, I get the error shown in the screenshot below. Courtesy of `ensure_dir_exists` in fs.lua. Ideally this stuff would go in `stdpath('data')` or `stdpath('cache')`

![screenshot-2024-10-10T09:15:53-07:00](https://github.com/user-attachments/assets/75f87dd6-bb26-42a5-9296-323be5d2b684)

gorillamoe:
Yes, we used `stdpath` before, but it was causing issues with windows filepaths :cry: .

Seems that the returned value from that does not have the correct path separator for windows :thinking: so this was an attempt to work around that.

Strage thing you even have that error message, are you using scripts?

willruggiano:
I am not, no. Didn't even know about them :D

gorillamoe:
Ah, shit the "fix" in main won't work either, because everything is happening in the tmp dir in kulala atm.. I have to overhaul that part completely I guess.

PR #284, opened 2024-10-10
fix(tmp): use `stdpath('cache')` for tmp files

gorillamoe:
This should fix: #282

Maybe @Grueslayer can check if this also works on windows.

Basically put: if you can still see request responses it should work.


gorillamoe:
Maybe also @willruggiano can check if this fixes the issue.

willruggiano:
Yeah this resolves my original error. I've got a new error now which I'll create a separate issue for

gorillamoe:
> Yeah this resolves my original error. I've got a new error now which I'll create a separate issue for

Should both work in this branch right now, fix for scratchpad is also in main, but I'll wait for having these fs changes tested on windows.

I'll try to have them tested in a docker container image from windows, let's see if I get this working :crossed_fingers:

gorillamoe:
> I'll try to have them tested in a docker container image from windows, let's see if I get this working 🤞

this is really complicated to get to work if you are used to the way linux / mac does the things.. so you need docker desktop and it needs to be running (neglecting the fact that docker desktop bugs me about creating an account every time I boot it up) and then you can run docker commands in poershell, but only if you have setup wsl2 correcly for which you need hyper-v enabled.. wtf?

I guess we get there, why is it like this? I mean why do we even need to have use docker on windows to build a docker container with windows? I mean it defeats the whole purpose imho :thinking:

gorillamoe:
Quick update, got the container running with neovim and all stuff, but need to figure out a way to get the right luarocks version installed with busted in the right version 🤔

gorillamoe:
Now, that I got everything running in that windows container, I can throw it in the trash 🗑️, because GitHub doesn't support custom runners except ones that are based on Linux ..

I'm going to port this to a workflow file running windows-server2022.

PR #287, opened 2024-10-11
feat(parser): add a HTTPHeaders table tuned to a work with HTTP headers

vyskocilm:
1. it's used as regular lua table h[name] or h[name] = value works
2. names are case-insensitive
3. values can be "flattened", so a name, single-value is returned. This
   is handly for curl command line generation.


lua/kulala/parser/headers.lua
```diff
@@ -0,0 +1,85 @@
+--- This class is used to manage the HTTP headers
+--- HTTP header has a case-insensitive name - so class store it in a lower case
+--- and can have a multiple values - which are stored in a table
```
vyskocilm, line 3: The aim is to replace the `kulala/parser/utils.lua` code with this abstraction. I tried to provide a minimal API tailored to kulala's main job converting HTTP specs to curl command line.

At the same the result behaves almost like a normal Lua table, so it's easy to use.

lua/kulala/parser/headers.lua
```diff
@@ -0,0 +1,85 @@
...
+      end
+    end
+  end)
+  local idx = 0
+  ---@return integer|nil
+  ---@return string[]|nil
+  local next = function(_, _)
+    local key, value
+    repeat
+      _, key, value = coroutine.resume(co)
```
vyskocilm, line 74: Frankly I haven't found easier way how to flatten the `string => []string` to `string => string` than using coroutines. I really did not want to manage the state when traversing the `_headers` table and this is where coroutines helps.

lua/kulala/parser/headers.lua
```diff
@@ -0,0 +1,85 @@
...
+---@param key string
+---@return any
+function HTTPHeaders.__index(self, key)
+  return rawget(getmetatable(self), key) or rawget(self._headers, key:lower())
+end
+
+---@param self HTTPHeaders
+---@param key string
+---@param value string
+function HTTPHeaders.__newindex(self, key, value)
```
vyskocilm, line 49: This is an equivalent of https://developer.mozilla.org/en-US/docs/Web/API/Headers/append - dunno if kulala needs more functionality than appending. More methods can be added.

gorillamoe:
I think this looks really good, with a small nitpick. I need to have a way to check for the existance of a specific header with a specific value, but want to have both header name and header value treated case insensitive, e.g. so I can check for this:

```lua
headers.has("x-request-type", "graphql", case_sensitive)
```

So param 2 and 3 could be potentially nil in any combination, where I can also have something like this

```lua
headers.has("X-REQUEST-TYPE", nil, true)
```

Where it would look for an all uppercase header and does not care about its value.

I don't think we have a use-case for having a strict case sensitive check as of now, so you might want to skip that part, but I need a way to check for these headers with values.

vyskocilm:
@gorillamoe please check the latest change - table have two methods `has` and `has_exact`. Both can name or name/value pair.

gorillamoe:
I think that this is yet superseeded by other commits, but thanks for working on that!

PR #294, opened 2024-10-19
feat: set buftype of scratchpad to nofile

Foo-x:
Currently, if you try to quit vim after opening a scratchpad, you'll either get an error or be prompted to save.
Since scratchpads aren't meant to be saved, this behavior is weird.
This PR fixes the issue.


Grueslayer:
@Foo-x only a revert of #279 is needed. We are waiting for the return of LazyVims maintainer to fix it in his repo. Than we’ll revert #279 and everything should be fine

Foo-x:
Sorry, I'm new to this repo, so I didn't know about that context. I'll close this PR.

PR #299, opened 2024-10-27
fix(parser): handle very large bodies

ng-tony:
If you try to create a request with a large body the following error occurs.

```
Error executing vim.schedule lua callback: Vim:E903: Process failed to start: argument list
 too long: "/usr/bin/curl"                                                                 
stack traceback:                                                                           
        [C]: in function 'jobstart'                                                        
        ...ocal/share/nvim/lazy/kulala.nvim/lua/kulala/cmd/init.lua:88: in function 'run_pa
rser'                                                                                      
        ...local/share/nvim/lazy/kulala.nvim/lua/kulala/ui/init.lua:205: in function <...lo
cal/share/nvim/lazy/kulala.nvim/lua/kulala/ui/init.lua:194> 
```

This looks like it is due to a shell limitation.
This fix write the body to a temporary file and passed that to curl.

### DIFF
--- a/lua/kulala/parser/init.lua
+++ b/lua/kulala/parser/init.lua
@@ -711,8 +711,13 @@ M.parse = function(start_request_linenr)
         Logger.error("Failed to create a temporary file for the binary request body")
       end
     else
-      table.insert(res.cmd, "--data")
-      table.insert(res.cmd, res.body)
+      local tmp_file = FS.get_text_temp_file(res.body)
+      if tmp_file ~= nil then
+        table.insert(res.cmd, "--data")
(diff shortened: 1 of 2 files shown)


ng-tony:
Putting this on draft because I've come to realize that this patch has would affect the curl preview feature.

ng-tony:
ref: https://www.in-ulm.de/%7Emascheck/various/argmax/

gorillamoe:
Implemented in https://github.com/mistweaverco/kulala.nvim/pull/305

Thnaks for raising this up! <3

PR #310, opened 2024-11-02
fix: space character problem when parsing content-type

zruhao:
Solve the problem that the response cannot be formatted when the content-type contains spaces, such as content-type: application/json   ; charset=utf-8

### DIFF
--- a/lua/kulala/internal_processing/init.lua
+++ b/lua/kulala/internal_processing/init.lua
@@ -3,6 +3,7 @@ local FS = require("kulala.utils.fs")
 local GLOBALS = require("kulala.globals")
 local DB = require("kulala.db")
 local CONFIG = require("kulala.config")
+local STRING_UTILS = require("kulala.utils.string")
 local M = {}
 
 -- Function to access a nested key in a table dynamically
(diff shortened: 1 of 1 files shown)


Grueslayer:
@zruhao As your fix does not harm kulala, your example is not valid. Since the first RFC 822, there is no space allowed

```
Content-Type := type "/" subtype *[";" parameter] 

type :=          "application"     / "audio" 
          / "image"           / "message" 
          / "multipart"  / "text" 
          / "video"           / x-token 

x-token := <The two characters "X-" followed, with no 
           intervening white space, by any token> 

subtype := token 

parameter := attribute "=" value 

attribute := token 

value := token / quoted-string 

token := 1*<any CHAR except SPACE, CTLs, or tspecials> 

tspecials :=  "(" / ")" / "<" / ">" / "@"  ; Must be in 
           /  "," / ";" / ":" / "\" / <">  ; quoted-string, 
           /  "/" / "[" / "]" / "?" / "."  ; to use within 
           /  "="                        ; parameter values
```

gorillamoe (approved):

PR #312, opened 2024-11-07
fix: result panel didn't show after call

LintaoAmons:
How to reproduce the issue:
- call and get the result window
- a new buffer took over the window
- the following call won't be able to see the result window

kulala


gorillamoe:
Thanks for the fix! :heart_eyes:

gorillamoe (approved):

PR #318, opened 2024-11-21
fix: change units in stats to be millseconds

ng-tony:
Stats panel incorrectly shows `Total Time` units in milliseconds while showing seconds.

Instead of changing the units to seconds. I thought it would be more useful to show all units on the stats page as milliseconds.
Before:
image

After
image


gorillamoe:
Thanks 👍🏾 ❤️

PR #320, opened 2024-11-25
fix: Authorization headers not showing on kulala.inspect

estacioneto:
As the title says, it fixes the issue where the headers were not shown on `kulala.inspect`, which would make `kulala.run` not work inside the inspect buffer.

The problem was described at https://github.com/mistweaverco/kulala.nvim/pull/122

### DIFF
--- a/lua/kulala/parser/init.lua
+++ b/lua/kulala/parser/init.lua
@@ -512,6 +512,7 @@ end
 ---@field url_raw string -- The raw URL as it appears in the document
 ---@field url string -- The URL with variables and dynamic variables replaced
 ---@field headers table -- The headers with variables and dynamic variables replaced
+---@field headers_display table -- The headers with variables and dynamic variables replaced and sanitized
 ---@field headers_raw table -- The headers as they appear in the document
 ---@field body_raw string|nil -- The raw body as it appears in the document
 ---@field body_computed string|nil -- The computed body as sent by curl; with variables and dynamic variables replaced
@@ -536,6 +537,7 @@ function M.get_basic_request_data(start_request_linenr)
     url = "",
     url_raw = "",
     headers = {},
+    headers_display = {},
     headers_raw = {},
     body = nil,
     body_raw = nil,
@@ -630,6 +632,8 @@ M.parse = function(start_request_linenr)
   -- for non existing variables
   res.url, res.headers, res.body, res.body_display =
     replace_variables_in_url_headers_body(res, document_variables, env, has_pre_request_scripts)
+
+  res.headers_display = vim.deepcopy(res.headers)
 
   -- Merge headers from the $shared environment if it does not exist in the request
   -- this ensures that you can always override the headers in the request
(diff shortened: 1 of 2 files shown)


Grueslayer:
The Authorization header is only removed from the header table when it should be handled by Curl itself (like for BASIC-User-Password / NTLM ....) but for other cases where there is no additional handling needed (like BEARER or BASIC where the Username and Passwort was already Base64-encoded) it keeps in the header list and still must be forwarded to Curl as header.

So the solution needs to be a little bit different, I'll have a look at it in the next days.

estacioneto:
Hi, folks!

> _The Authorization header is only removed from the header table when it should be handled by Curl itself but for other cases where there is no additional handling needed **it keeps in the header list and still must be forwarded to Curl as header**._

It was not being kept in the headers list and that's how I thought we could solve it, by adding it as an authorization header like it was, with `-H`.

I'm not sure I understood why a different solution is needed, but of course feel free to change it and fix it in another way as long as the inspect window shows the authentication headers, which is important for me to be able to call `kulala.run` inside the inspect buffer.

Grueslayer:
Because the "written" header is mostly not what is really send.

E.g. You have user/passwort credentials like

- User = `MyUsername`
- Password = `Abc123`

you have to send a header: 

`Authorization: BASIC TXlVc2VybmFtZTpBYmMxMjM=`

**But** mostly you don't, you are writing:

`Authorization: BASIC MyUsername:Abc123`

and that is **not** send in this way. We remove that header and let create `curl` create the right one (Base64 encoded for Basic).
There are also other like `AWS`. These are totally different from just sending a plain header line. AWS will create two requests!

There are cases where no special treatment are necessary (like shown before, when user already has encoded the Basic authentication in Base64).

Kulala removes the Auth header only in those cases when a special treatment is implemented and handled by `curl` itself. 

So

1. We can leave the current written Auth header in the lua table
2. But we need to suppress the `-H` argument only in  those cases where we previously have removed the header line

So instead to remove the header line from the table, we have to set a local boolean (say `SkipAuthorizeHeader`) at those progam lines and in the loop where `Authorize` is been added to the curl argument list skip those action only when boolean variable is true.

Inspect shows now in all cases the previosly written line but still can be different whar curl is sending to the server!

estacioneto:
> _We remove that header and let curl create the right one..._

I understand it and that's why I didn't touch the lines above `res.headers[auth_header_name] = nil`. This way the authorization is still passed to curl the right way.

> _But we need to suppress the -H argument only in those cases where we previously have removed the header line_

Which is what `if key ~= auth_header_name then` should be doing.

But I understand I might still be missing a scenario and that's why I don't fully understand the difference, but I'll apply the changes you suggested :)

gorillamoe:
Hey there, if I understand this correctly, this could also go the way we do with a separate variable used for display, like we already do with body, right? I think this should be the way to go.

I think I can prepare something.

gorillamoe:
@estacioneto, can you check the current state, if this is what you have desired?

Now, it shows all headers (including the auth headers) on inpsect.

estacioneto:
@gorillamoe it works for me! I'll close the PR.

Thank you so much!

Issue #66, opened 2024-07-22
Compatibility with vscode Rest-Client 


gorillamoe:
Make it work with https://github.com/Huachao/vscode-restclient

Multiline url
- [ ] Shorthand form of request (without GET)
- [x] GQL
- [x] Request Variables (see https://github.com/mistweaverco/kulala.nvim/discussions/61)

X-ref: https://github.com/mistweaverco/kulala-fmt/issues/2

gorillamoe:
I think this is way to big for one issue. Implementation of Request Variables is done. I'm going to close this issue and we work out what features wwe really need to adapt :see_no_evil:.

igorlfs:
Hello!

> Multiline url

Is that still on the roadmap? Is it being tracked somewhere?

gorillamoe:
What s missing for you? @igorlfs

Grueslayer:
This has been discussed and implemented by 

https://github.com/Huachao/vscode-restclient/issues/16

And I follow the thoughts about breaking highlighting and using word wrap.

gorillamoe:
> Shorthand form of request (without GET)

I don't think we will support his anytime soon, so I'm closing this.

PR #329, opened 2024-12-18
Query parameters can be placed in a new line

Grueslayer:
Adds multline url feature identical to vscode-rest-client:

Every new line starting with

- `?`
- `&`

following the request line is added to the url.


Grueslayer:
Adds multiline url feature to #66

gorillamoe (approved):

PR #336, opened 2025-01-21
fix(format): fix function format params

edte:
1. When setting a custom formatter, an error will be reported for base_table that does not exist.

```
                ["text/tsv"] = {
                    ft = "tsv",
                    formatter = function(body)
                        return body:upper()
                    end,
                },
```

2. When using a custom formatter to process the returned buffer, if the autocmd of filetype is set, the first trigger will be an empty buffer.


gorillamoe (approved):

gorillamoe:
Thanks for your contribution! :heart:

PR #338, opened 2025-01-26
Add verbose mode for Curl output Issue#331

YaroSpace:
Added verbose mode or `default_view` config option.
Updated winbar for output buffer to include `Verbose`
Added tests for verbose mode and some generic test helpers.
Updated docs.
Some minor refactoring on the touched code.

Redirect of verbose output to a file is not implemented for now.  Can do if needed.


lua/kulala/parser/init.lua
```diff
@@ -693,6 +693,7 @@ M.parse = function(start_request_linenr)
   table.insert(res.cmd, "@" .. CURL_FORMAT_FILE)
   table.insert(res.cmd, "-X")
   table.insert(res.cmd, res.method)
+  table.insert(res.cmd, "-v")
```
gorillamoe, line 696: Haven't tested it yet, but doesn't this cause issues for non-verbose requests?

gorillamoe:
Linter is also failing.

lua/kulala/parser/init.lua
YaroSpace: I have not noticed any.  The only difference for verbose requests is that curl pipes verbose output to stderr.  Non verbose  requests just ignore it.

Issue #337, opened 2025-01-25
Parsing errors when executing `run_all()`


YaroSpace:
I am getting parser errors when trying to run multiple requests in a file with `run_all()` when default_view is `float`.
I found that the document gets parsed with `M.get_document()` several time before each request.  Once the first request goes through, the float opens and the current buffer changes, so all  subsequent calls to `get_document` return nil, because it is now trying to parse the float.

I have fixed it locally.  Will submit a PR to have a look.

On a connected issue, how are multiple requests supposed to be visualized?  At the moment the buffer is just overwritten with the last request.

gorillamoe:
> On a connected issue, how are multiple requests supposed to be visualized? At the moment the buffer is just overwritten with the last request.

I don't think of really visualizing each request when run_all, but mostly of just show me the last request, becuase it's mostly a series of requests where I just care about the last result.

But maybe my use-case for that is too narrow.

What's your take on that @Grueslayer?

PR #342, opened 2025-01-27
Fix: parsing errors when doing run_all() #337

YaroSpace:
Issue #337

1. Fix for parsing errors when doing `run_all()` due to interfering floating window.  As a side effect, the document is now only parsed once for one request instead of 3 times as it had been before.
2. Fix for virtual text icons that fail for the same reason.
3. Minor fix for `DB.update()` that fails if the current buffer has changed.
4. Added some error handling and messaging, when parsing jobs fail.
5. Added stubs for jobstart/system/curl to allow functional testing.
6. Added functional tests: the current set should cover most of the functionality: UI, parsing, JS pre/post/inline processing, input/notifications, document/env.json/dynamic variables.

Please do not be overwhelmed by number of files changed.  The majority are specs.

The problem was in using Nvim functions that rely on `current buffer`, e.g. get_lines, get_cursor, get_buffer, get_namespace expanding file names based on current buffer, etc.  So, when a float opens, the current buffer changes and things start to fail around the place.
I removed the coupling by introducing `DB.current_buffer`, which holds the id of the buffer for current request.


gorillamoe:
Thanks for your hard work and awesome 😎👍🏾 contributions ❤️.

gorillamoe:
This PR breaks something like this, which works on main:

```http
# Setting the environment variables to be used in the next request.
# Any external command can be used to set the environment variables.
# The command should output the environment variable as string.
# @env-stdin-cmd JWT_CONTEXT jq -rc '.json.token | gsub("-";"+") | gsub("_";"/") | split(".") | .[1] | @base64d | fromjson | .ctx'
POST https://httpbin.org/post HTTP/1.1
Accept: application/json

{
  "token": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ0eXBlIjoiZ29yaWxsYSIsIm5hbWUiOiJHb3JpbGxhIE1vZSIsImN0eCI6IlNvbWUgY29udGV4dCIsIndlYnNpdGUiOiJodHRwczovL2dvcmlsbGEubW9lIn0.YmEG9bOo1o9opeWnCsfW621A-sB_5WXSBI2FjtvwXlk"
}

###

POST https://httpbin.org/post HTTP/1.1
Accept: application/json

{
  "success": true,
  "context": "{{JWT_CONTEXT}}"
}
```

YaroSpace:
Let me have a look. 

Can you send me an http file with a bunch of different requests/usecases, so I can run the tests against it?

gorillamoe:
[http-examples.zip](https://github.com/user-attachments/files/18571990/http-examples.zip)

Attached you'll find a collection of requests I usually tend to test.

For the scripts to work, you need to `cd` into the `scripts` dir and do a `npm install` first.

Then cd into `one` and open one of the `.http` files.

YaroSpace:
So, I have checked a few times.  There was a typo in token - `foo` in the http file you sent me at 
```
# @name REQUEST_TWO
POST https://httpbin.org/post HTTP/1.1
accept: application/json
content-type: application/json

{
  "token": "{{REQUEST_ONE.response.body.$.json.foobar}}",
  "gorilla": "{{RANDOM_DATE}}"
}
```
Otherwise, all requests go through without errors.

gorillamoe:
Yes, the requests work, but the global variables are not getting filled, when requests are made. I'll try to make a short screencast about that for you.

gorillamoe:
Nevermind, maybe I was just too tired when checking. LGTM. Thanks for the rework!

YaroSpace:
> Yes, the requests work, but the global variables are not getting filled, when requests are made. I'll try to make a short screencast about that for you.

That would be great.  There is definitely some bug somewhere in saving variables, which I am having hard time to track down.  The requests and variables seem to be fine on both this branch and main in split view, but do fail with float in both branches.  On main it fails with the same error I had from the start, which was the reason for this PR, but I thought that I fixed it.

gorillamoe:
first example in advanced.http in the attached .zip of my previous post already fails in the reworked branch.

[Screencast from 2025-01-29 19-33-11.webm](https://github.com/user-attachments/assets/94f7af6b-f6dd-47a9-bccb-782b8cead86e)

YaroSpace:
Yep, found the problem and finally fixed it.  It was a hell of a bug hunt).

gorillamoe:
@YaroSpace, you did not push it yet, right?

YaroSpace:
No, not yet.  I have updated it and added some functional specs that use npm, so some specs are failing .  Just dealing with that.

YaroSpace:
> @YaroSpace, you did not push it yet, right?

Done now.

YaroSpace:
That was the final force push.  I was eager to finish with specs, so to be safe when adding new features.  The current set

gorillamoe:
Looks great and I could not find a failing query! Thanks sooooo much :heart: .

gorillamoe (approved):

YaroSpace:
It was fun! 👌

Issue #325, opened 2024-12-12
Feature Idea: Send Request for URL at cursor in any buffer


kevintraver:
It would be useful to have an additional function/method that could work in any buffer, not just _http_ filetype

Here is an example use case:

Imagine you are in a JSON file, and one of the key values is a url. You could place the cursor on the URL and send a request with a single keymap.

This could always assume the use of the GET method.

vdegenne:
Agreed, for now you can use:

```lua
vim.keymap.set("n", "<leader>r", function()
  -- Get the current line the cursor is on
  local line = vim.fn.getline(".")

  -- Define an updated regex pattern for matching HTTP/HTTPS URLs with optional ports
  local url_pattern = [[https?://[%w-_%.%d:/?&=]+]]

  -- Find the URL in the current line
  local url_under_cursor = string.match(line, url_pattern)

  -- If a URL was found, proceed with the curl request
  if url_under_cursor then
    local result = require("plenary.job")
      :new({
        command = "curl",
        args = { "--silent", url_under_cursor },
      })
      :sync()

    if result then
      Snacks.notify.info(result, { ft = "json" })
    end
  else
    Snacks.notify.warn("No URL found in the current line")
  end
end, { noremap = true, silent = true })
```

gorillamoe:
Basically it's not possible to reduce complexity in our code atm, but you can have requests run in markdown files, if that helps.

gorillamoe:
The current implementation allows you to have fenced code blocks with http syntax in it which kulala can parse.

Assume this .py file:

```python
"""

```http
# Setting the environment variables to be used in the next request.
# Any external command can be used to set the environment variables.
# The command should output the environment variable as string.
# @env-stdin-cmd JWT_CONTEXT jq -rc '.json.token | gsub("-";"+") | gsub("_";"/") | split(".") | .[1] | @base64d | fromjson | .ctx'
POST https://httpbin.org/post HTTP/1.1
Content-Type: application/json
Accept: application/json

{
[...]

PR #351, opened 2025-02-06
Send request under cursor or visual selection #325

YaroSpace:
Added support for running request on the current line or in the visual selection for non-http/rest files.

When you run `require('kulala').run() ` or press the corresponding keymap, the parser will work in this sequence:
1. If http/rest file - parse the whole document, run the request within the current line.
2. If non-http/rest - 
  a) look for fenced block  
  b) use current visual selection  
  c) use current line
pass to parser and run the request within current line.

All common comment characters are stripped, so will work for requests that are commented out in source code.


gorillamoe (approved):

gorillamoe:
Awesome work :muscle: :+1: Thanks again for a new wonderful feature!

PR #352, opened 2025-02-06
Fix: to allow run_all() for visually selected lines (#325)

YaroSpace:
MInor update for #351  to allow `run_allI()` for visual selection.


gorillamoe:
> On a related matter, maybe it is a good idea to include the ftplugin file with keymaps for http/rest and also add default keymaps to the config

I think that might break LazyVim Extras integration, no?

> so that they can be switched on/off with one option

I think that would be indeed a good addition.

PR #370, opened 2025-02-14
Fix: Send one-off request async (#368)

### LINKED ISSUE #368: Requests are no longer run asynchronously
Hey there!

I've noticed the plugin has gone through quite a few changes lately. That's nice!

However, the recent changes also made requests run synchronously. Is this intentional? If so, is there any way to make get around it?

I'm asking because my main use case for Kulala is _debugging_ requests. With the recent changes, that's no longer possible: Kulala completely halts nvim's UI, making it impossible to interact with nvim-dap (the only way to escape if a breakpoint is hit is to use `<C-c>`, which cause the request to fail).

I thought about creating a discussion instead, but I have the feeling the introduction of this behavior was unintentional.

If anyone else runs into this, my current workaround to handle debugging is to use the [copy as curl](https://neovim.getkulala.net/docs/getting-started/example-configuration#copy-as-curl) function, and then running the request via curl on another terminal.

YaroSpace:
Resolves #368 

Problem:
- a regression introduced in #346 when removing `vim.fn.jobstart` in favour of `vim.system`.   Single requests when run with `vim.system` are queued with `wait()`.

Fix: 

- Skip `wait()` for single requests.

Change:
- added config option `request_timeout = nil` by default, to set max timeout.

### DIFF
--- a/lua/kulala/cmd/init.lua
+++ b/lua/kulala/cmd/init.lua
@@ -1,4 +1,5 @@
 local GLOBALS = require("kulala.globals")
+local CONFIG = require("kulala.config")
 local Fs = require("kulala.utils.fs")
 local PARSER = require("kulala.parser")
 local EXT_PROCESSING = require("kulala.external_processing")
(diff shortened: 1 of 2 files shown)


gorillamoe (approved):

gorillamoe:
This is huge work! Thanks so much :heart: !

PR #371, opened 2025-02-14
Fix: allow async for run_all() (#368)

YaroSpace:
This is a follow up for #368 to allow async for `run_all()` requests.


gorillamoe:
Haven't checked yet, but does this still work with the `@prompt` directive?

YaroSpace:
Yes, it does.  The requests are still run sequentially, it is just the next one is fired off in the `on_exit` callback instead of after `.wait()`, so this does not block the UI during the actual request.

gorillamoe:
This is a real huge awesome pr. This was bugging me lately and you fixed it. Thanks so much!

PR #376, opened 2025-02-15
Feat: add request/response history

YaroSpace:
Feature:
- Store responses history during a Neovim session
- Allow to scroll backwards and forward through responses

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ http-client.env.json
 .tests
 .nvim
 package-lock.json
+todo.md


gorillamoe (approved):

YaroSpace:
@gorillamoe you have pulled in all the changes with the last PR.

LintaoAmons:
How to use this feature, I can only see there is one file change during this PR?

YaroSpace:
Very simple - use [ ] to scroll through history.

The changes are in the previous PR.

LintaoAmons:
Sorry, maybe I missed something...
But I want to ask how to open the floating window that you are showing in the description, the window with shortcut hints at the top

YaroSpace:
You should be seeing it in the request output window, does not matter if it is split or float.

gorillamoe:
I think @LintaoAmons wants this

https://neovim.getkulala.net/docs/getting-started/configuration-options/#display_mode and this
https://neovim.getkulala.net/docs/getting-started/configuration-options#winbar

LintaoAmons:
Thanks! 
I would like to say that winbar should default to true

LintaoAmons:
image
Is it possible to add one more field to display the sent datetime  so the user can easily differentiate the responses

YaroSpace:
Sure!

gorillamoe:
> > I would like to say that winbar should default to true
> 
> I think so too. @gorillamoe what do you think?

I'm okay with that. Should be a major release then, because it's "breaking" the current setup. We might also remove `@file-to-variable` then.

PR #377, opened 2025-02-15
Code style

YaroSpace:


gorillamoe (approved):

gorillamoe (approved):
If I understood that all correctly, not only codestyle was changed, but also some refactoring has been done and also some expcted test values have been changed, but looks really good.

PR #382, opened 2025-02-20
Docs UI

### LINKED ISSUE #385: how to disable this header
After I updated to v4.11.0, this header appeared, so that body can't be formatted normally.

Image

config
```
    {
        'mistweaverco/kulala.nvim',
        ft = "http",
        opts = {
            default_view = "body",
            display_mode = "float",
            winbar = false,
            contenttypes = {
                ["application/csv"] = {
                    ft = "csv",
                },
                ["text/csv"] = {
                    ft = "csv",
                },
                ["text/tsv"] = {
                    ft = "tsv",
                    formatter = function(body)
                        return body
                    end,
                    pathresolver = function(body, path)
                        return body
                    end,
                },
            },
        },
    },

```

YaroSpace:
## Some minor fixes, UI/UX and docs updates

Resolves #385 #386 

- add time to response summary
- silently skip commented out URLs in request blocks
- allow use of variables in gRPC requests
- expand paths in `grpc-import` flags
- set `winbar = true` as default
- add `show_request_summary = true` option
- remove support for @file-to-variable
- group UI options under `config.ui`

==
- add response summary spec, redirect-to-file spec
- update docs


gorillamoe:
You also removed the docs for file-to-variables, yes? Can you also remove that code that belongs to file-to-variable?

YaroSpace:
Oh, sorry! I did after your post on discord and forgot to mention in the description. 

At the moment it is just a few lines and I thought we would leave it for backwards compatibility for a few releases, but tell me what's best.

gorillamoe:
I would like it to be removed, because we're going to bump Kulala to v5 then, which is a new major release and then we can have these breaking changes. Otherwise we would have to bump to v5 for the other breaking changes, then after little time do another major bump to v6 just for removing the file-to-variable.

Thanks for your continous effort to make Kulala the best http client experience in Neovim! <3

gorillamoe (approved):

PR #384, opened 2025-02-20
fix(curl parser): multiline body (#383)

### LINKED ISSUE #383: Paste from cURL with multiple --data parameters
Paste from curl does not work with multiple `--data` parameters, it only prints the last one. 
`--data-urlencode` does not seem to work at all.

```http
# curl  --request GET 'https://some-url' --header 'Content-Type: application/x-www-form-urlencoded' --data 'client_id=guid' --data 'scope=api://guid' 
GET https://login.microsoftonline.com/<id>/oauth2/v2.0/token
content-type: application/x-www-form-urlencoded

scope=api://guid
```

```http
# curl  --request GET 'https://some-url' --header 'Content-Type: application/x-www-form-urlencoded' --data-urlencode 'client_id=guid' --data-urlencode 'scope=api://guid' --data-urlencode 'client_secret=secret_value'
GET https://login.microsoftonline.com/<id>/oauth2/v2.0/token
content-type: application/x-www-form-urlencoded
```

bosvik:
Added match for "-data" to also iclude "--data-urlencode".
Iter body table and print each line.


YaroSpace:
Oh, thanks for this. This needed to be fixed. Will have a look today.

gorillamoe:
Wow, thanks!

gorillamoe (dismissed):

bosvik:
Thanks for making this kickass plugin!

bosvik:
@gorillamoe sorry I completely missed you had made some comments 😵 .

YaroSpace:
Hey @bosvik thank you for your input!  Would you please squash the commits into one and rename it to `fix(curl parser): multiline body (#383)`

YaroSpace (changes requested):
Hey @bosvik thank you for your input! 
Would you please squash the commits into one and rename it to `fix(curl parser): multiline body (#383)`?

gorillamoe:
> Hey @bosvik thank you for your input! Would you please squash the commits into one and rename it to `fix(curl parser): multiline body (#383)`?

No need to squash, that is automatically done when we accept the PR.

badloop:
Does this actually work?  I'm currently attempting to do this same thing for authentication and there seems to be no way to replicate in the .http format what curl uses `--data-urlencode` for

YaroSpace:
This fix was for the `copy from curl` command.
What is it that you are trying to do?

YaroSpace:
Maybe this helps https://neovim.getkulala.net/docs/usage/sending-form-data?

YaroSpace:
You can try doing
```http
# @curl-data-urlencode` on your request
POST url.com
```

badloop:
> You can try doing
> 
> ```httpspec
> # @curl-data-urlencode
> POST url.com
> ```

This just adds that flag to the request.  What I am specifically trying to do is authenticate with salesforce which uses flags on the curl like this:
`--data-urlencode 'grant_type=client_credentials'`

YaroSpace:
Ah, sorry. I pushed that fix to develop branch.

Can you please update from develop and try again?

badloop:
Still no dice... its just parsing the variables into a payload.
image
image

YaroSpace:
You should get rid of the header, `Content-Type: application/x-www-form-urlencoded` - implies that the payload is already encoded.
You also need `&` to separate the `name=value` pairs

YaroSpace:
image

The payload is shown as it is passed to curl, before it is encoded by it.  But the final result is encoded.

badloop:
OK I think this is me just misunderstanding the setup.... i was assuming that the query parameters could be defined like this but it appears they need to be in the URL itself to work correctly here.  Thanks for your help!

YaroSpace:
For Oauth2 grant types Authorization Code and Implicit they are indeed supplied in the url.
You may want to give Kulala's Oauth2 support a try.

PR #397, opened 2025-02-24
Feat: replace vars in external json

YaroSpace:
## Features:
- add support for variables in external `json` files

## Fixes:
- add gRPCurl check to checkhealth

### DIFF
--- a/lua/kulala/health.lua
+++ b/lua/kulala/health.lua
@@ -6,7 +6,7 @@ local health = vim.health
 
 local M = {}
 
-local function check_executalbe(name, path)
+local function check_executable(name, path)
   if FS.command_exists(path) then
     path = FS.command_path(path)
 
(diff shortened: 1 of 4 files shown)


gorillamoe:
doesn't this cause an exception when binary files are used?

YaroSpace:
It should not. 
The variables are replaced only in files with .json extension, which should not have any binary data in them by definition, unless it is encoded. The most common encoding of base64 does not contain {} characters, so that will not interfere with variables syntax.

gorillamoe:
> It should not. The variables are replaced only in files with .json extension, which should not have any binary data in them by definition, unless it is encoded. The most common encoding of base64 does not contain {} characters, so that will not interfere with variables syntax.

Ah, I missed that restriction to json :see_no_evil:

gorillamoe (approved):

PR #402, opened 2025-02-25
fix: use nowait for keymaps

igorlfs:
Hello!

With the introduction of `[` and `]` as means to navigate the request history, one thing that's bugging me out is the delay. Since there are default mappings that start with `[` and `]` (`[[` and `]]` respectfully), neovim waits `'timeoutlen'` for another key before executing the initial action.

To handle that, we can just make all mappings use the `nowait` option. From my testing, this does not interfere with any other default keymaps.


YaroSpace:
I agree, it makes sense.  Thank you!

YaroSpace (approved):

PR #409, opened 2025-03-10
Open new win that respects user split settings

dakennguyen:
With this PR, we don't force split to be `right` or `below`, but it will depend on `splitright` `splitbelow` settings

### DIFF
--- a/lua/kulala/ui/init.lua
+++ b/lua/kulala/ui/init.lua
@@ -124,7 +124,7 @@ local function open_kulala_window(buf)
       style = "minimal",
     }
   else
-    win_config = { split = config.split_direction == "vertical" and "right" or "below", win = request_win }
+    win_config = { vertical = config.split_direction == "vertical", win = request_win }
   end
 
   win = vim.api.nvim_open_win(buf, true, win_config)


YaroSpace:
I found that using `vertical` and `sb|spr` works unreliably to open a split below.
I guess we could do something like:
```lua
  win_config = { 
split = config.split_direction and (config.split_direciton == "vertical" and "right" or "below") or nil,
vertical = not config.split_direction and true or nil 
win = request_win
}

```

dakennguyen:
could you let me know in what scenario it works unreliably exactly?

For my PR, I tested these cases:
1. `config.split_direction = 'vertical'` , `:set splitright` => open a vertical split to the right
2. `config.split_direction = 'vertical'` , `:set nosplitright` => open a vertical split to the left
3. `config.split_direction = 'horizontal'` , `:set splitbelow` => open a horizontal split below
4. `config.split_direction = 'horizontal'` , `:set nosplitbelow` => open a horizontal split above

For the code snippet above ,`vertical = not config.split_direction and true or nil` will result to `nil` if my `config.split_direction == 'vertical'`, and I can't have the option to open a vertical split to the left, either.

YaroSpace:
It was a while ago, so I do not remember what the issue was exactly.
In any case, I agree that your way gives more flexibility to users and I have tested your code and it works fine.
Thank you for your input!

gorillamoe:
Is this a breaking change? I want the users still to be able to have kulala act different than their general neovim config. Is this still possible?

YaroSpace:
I was just about to say that.   It is not breaking, but may require users to set the `sb|spr` option if it is not set already, otherwise the split may unexpectedly open above|left, instead of right|below.  We need a way to preserve the accustomed behaviour. 
I will think of a way.

dakennguyen:
that would be nice if user can set split position regardless of vim settings, but I think on default, position based on vim settings should be the expected behavior. 

This change could impact users with `nosplitright` or `nosplitbelow`, as the window placement would differ from their current setup. But I doubt that many users would prefer Kulala to behave differently from their usual `:vsplit/:split` behavior

YaroSpace:
Could not update this PR directly, so created #417.

YaroSpace:
Superseded by #417

PR #414, opened 2025-03-11
fix: use screen size instead of win size for float

dakennguyen:
`nvim_win_get_width` and `nvim_win_get_height` returns the size of the current window, so if we have many windows in 1 tab, the opening float would be really small. I think we should use size of the screen instead
image


YaroSpace:
I agree. I meant to fix that a while ago.

gorillamoe (approved):

PR #417, opened 2025-03-15
Feat: add win_opts config option

YaroSpace:
Add an option to override default UI window settings. 
Closes #409

### DIFF
--- a/lua/kulala/ui/init.lua
+++ b/lua/kulala/ui/init.lua
@@ -127,6 +127,7 @@ local function open_kulala_window(buf)
     win_config = { split = config.split_direction == "vertical" and "right" or "below", win = request_win }
   end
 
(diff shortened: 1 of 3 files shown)


gorillamoe (approved):

lua/kulala/ui/init.lua
```diff
@@ -127,6 +127,7 @@ local function open_kulala_window(buf)
     win_config = { split = config.split_direction == "vertical" and "right" or "below", win = request_win }
   end
 
+  vim.tbl_extend("force", win_config, config.ui.win_opts or {})
```
dakennguyen, line 130:
```lua
win_config = vim.tbl_extend("force", win_config, config.ui.win_opts or {})
```
I think we have assign the returned value here

YaroSpace: You are absolutely right! Good catch!

PR #427, opened 2025-03-19
feat(doc): add runnable http demo

LintaoAmons:


gorillamoe:
Thanks for you continuous efforts to make Kulala really one of the best rest clients out there!

I would really like this to be part of the docs, can we maybe rework that PR.

YaroSpace:
Yep, I will do that.

YaroSpace (approved):
Thank you for your input.  I have integrated your demos into the docs.

PR #429, opened 2025-03-19
Feat: add news hint

YaroSpace:
Feat: add news hint, that shows up on version change

image


gorillamoe (approved):
What a really, really nice addition!

PR #430, opened 2025-03-23
Feat: $auth.token for Oauth2 support

YaroSpace:
Feat: $auth.token for Oauth2 support

Fully compliant with IntelliJ Spec.

Specify the auth parameters in the `http-client.env.json` under the desired environment:
You can then use `{{$auth.token('auth-id')}}` variable in the request Authorization header or in query parameters.

1.  All Grant Types are supported.
2.  A browser will be opened to perform user consent with the auth provider.  The browser will be redirected to provided `redirect_url`.  If the url is `localhost|127.0.0.1` the code will be received automatically, otherwise  you will need to manually copy the `code|access_token` from the browser url address pane into Kulala prompt.
3. Exchange of code for token and subsequent token refreshes will be done automatically.
4. Device token is copied into clipboard to paste into the verification form.
5. Custom params can be supplied through `Custom Request Parameters` key.
6. JWT is generated from `JWT` params.

Invoke Auth Manager with `<leader>Ru` to manage Auth configurations.


gorillamoe:
As we already have nodejs somehow integrated in our setup, we could also spin up a local webserver and intercept that code via the webserver, but that might be just overkill for the first iteration.

YaroSpace:
> As we already have nodejs somehow integrated in our setup, we could also spin up a local webserver and intercept that code via the webserver, but that might be just overkill for the first iteration.

Unfortunately, that would not work.  In order to perform the user consent, you have to open a link with a browser that support js.  It is then this browser gets redirected to the redirect_url with the code in the fragment.  We do not actually need a working endpoint, just to extract the code from redirect url or headers.

IntelliJ, Bruno, Postman are all using their own browser to do that.  They also provide an option to use system browser to enter credentials, but then redirect to their browser anyway.

I will try to use selenium and see if that works out.

gorillamoe:
> Unfortunately, that would not work. In order to perform the user consent, you have to open a link with a browser that supports js. It is then this browser gets redirected to the redirect_url with the code in the fragment. We do not actually need a working endpoint, just to extract the code from redirect url or headers.

I know, because I'm doing some Oauth stuff at work, but we just spin up a local webserver and for authentication, we provide a redirect url of let's say http://localhost:8488/auth_callback

and there we can intercept that code.

YaroSpace:
> > Unfortunately, that would not work. In order to perform the user consent, you have to open a link with a browser that supports js. It is then this browser gets redirected to the redirect_url with the code in the fragment. We do not actually need a working endpoint, just to extract the code from redirect url or headers.
> 
> I know, because I'm doing some Oauth stuff at work, but we just spin up a local webserver and for authentication, we provide a redirect url of let's say http://localhost:8488/auth_callback
> 
> and there we can intercept that code.

I mean it would not work without a browser, but intercepting the redirect with local server will of course work, I thought of that too.
Although, that would require to have control over the auth provider config to whitelist localhost redirects.  I am not sure how feasible that is for people, besides some auth providers specifically do not allow localhosts and some require https.

YaroSpace:
@gorillamoe implemented your idea.  Worked out quite nicely.

jeffglover:
I did some testing and got it working with Implicit. I'm came across 2 issues

- [This](https://github.com/mistweaverco/kulala.nvim/pull/430/files#diff-90fb4327bce47b04adb2a5458506e20ce7b87658ec5aa4ef0aee0d72b93e5102R113) requires `http-client.private.env.json`. But when it saves auth, will only try to save to `http-client.env.json`
- Looking forward to [state_parameter_passthrough_value option as you mentioned](https://github.com/mistweaverco/kulala.nvim/pull/430/files#diff-d635261534db531ac3f714aede46d4275001e01b101b56f93dde0127f17c69aeR206). Hardcoded what I needed for now.

Working config:

```json
    "Security": {
      "Auth": {
        "auth-id": {
          "Type": "OAuth2",
          "Redirect URL": "http://localhost:3000",
          "Grant Type": "Implicit",
          "Client ID": "REDACTED",
          "Auth URL": "REDACTED/oauth2/v2.0/authorize",
          "Scope": "openid https://REDACTED"
        }
      }
    }
```

Thanks for working on this, I'll keep testing as needed.

YaroSpace:
> I did some testing and got it working with Implicit. I came across 2 issues
> 
> * [This](https://github.com/mistweaverco/kulala.nvim/pull/430/files#diff-90fb4327bce47b04adb2a5458506e20ce7b87658ec5aa4ef0aee0d72b93e5102R113) requires `http-client.private.env.json`. But when it saves auth, will only try to save to `http-client.env.json`

When the env files are read they are merged into one table , both public and private.
Later, only `http-client.env.json` gets updated and if it contains sensitive fields that are also present in private.env, they will be sanitized from it.

YaroSpace:
> * Looking forward to [state_parameter_passthrough_value option as you mentioned]
Custom params can be supplied through `Custom Request Parameters` key

YaroSpace:
@jeffglover @e99train  All features in place.  Specs and docs left.
Looking forward for your feedback.

jeffglover:
Added `Custom Request Parameters` to my config and all works flawlessly, no more local changes required.

RE: env config files. I reread the docs and now better understand the idea behind .private. I'm new to kulala coming from rest.nvim -> hurl.nvim -> kulala.nvim.

e99train:
This is really great so far!

I have encountered a few issues while testing it out:

- Every time it tries to get a new token, it seems to reformat my http-client.env.json file to be completely flattened (all in a single line)
- The project I'm working on does not currently require the "Scope" property (we are using authorization code grant type so maybe we're not properly following spec or something, I haven't looked into it much yet), but if I leave the scope property out of the env, it does not make the request to get the token
- It seems like the redirect url might not be getting url encoded because it kept failing to redirect correctly until I entered the redirect url into the env already url encoded ("https%3A%2F%2Flocalhost%3A7179" instead of "https://localhost:7179")
- It would also be nice to be able to manually cancel the attempt to get a token if I see that it will fail instead of waiting for it to time out or come back with an error as it does block the nvim ui for me while it's waiting

Let me know if you need any more information regarding these issues, I'm happy to continue testing this out!

e99train:
After testing a little more, I don't think the redirect url encoding point is actually an issue, but I am having some issues with the redirect url in general. I will update if I find out more about this

e99train:
Ok it seems like the redirect url issues were on my end, not a problem with this plugin. I edited the plugin code locally so it wouldn't require scope and I am now able to get to the point where it does call the redirect url, but it is not currently retrieving the code even though it's going to localhost. Could this have something to do with the "Scope" field?

YaroSpace:
> * Every time it tries to get a new token, it seems to reformat my http-client.env.json file to be completely flattened (all in a single line)
Fixed that, but you will need `jq` installed.

> * The project I'm working on does not currently require the "Scope" property (we are using authorization code grant type so maybe we're not properly following spec or something, I haven't looked into it much yet), but if I leave the scope property out of the env, it does not make the request to get the token

`Scope` is indeed part of Oauth2, but I have made it optional.

> 
> EDIT: C-c does cancel the attempt to get the token and the redirect url issues were on my end
Yes, that's the intended way.

e99train:
Thank you for making those changes! There was one error on line 328 of oauth.lua when setting the body. Without the "Scope" value defined, it would set the body to nil I think, but I was able to fix it locally and was able to get it all working perfectly.

The fix I made locally was to change this line:
```
body = config["Scope"] and body .. "&scope=" .. config["Scope"]
```

to this line:
```
body = config["Scope"] and body .. "&scope=" .. config["Scope"] or body
```

YaroSpace:
My bad)  Rushed to make changes.

e99train:
Sorry to keep adding more things, but I just realized that the retrieved tokens are being saved to the http-client.env.json file. Would it be possible to have those be saved to the http-client.private.env.json file instead of the regular one?
This happens when I paste the code in so I'm not sure if the same thing happens when redirecting to localhost and auto detecting the token

YaroSpace:
Oh, please add as much as you can)
I am very glad to have feedback!

That's a very good point about the env file. I was actually thinking the same.  Tokens can certainly go to private.env, but what about the other params like expiry, scope, audience, etc. that get returned with tokens?

I will move them to private.env for now too, but maybe a better idea comes up.

e99train:
Personally, I would still prefer to have those other params be saved to the private env as well just so I can avoid those values being constantly changed in git.

But I agree those aren't necessarily private values as much as they are just specific to the local environment, so I am also open to other ideas.

gorillamoe (approved):

PR #432, opened 2025-03-24
feat(virt_text): use 'combine' to blend with cursorline

iovis:
Virtual text when used with `cursorline` overrides the BG color of the cursorline, using `hl_mode = "combine"` makes it so they blend together.

Before:
image

After:
image


YaroSpace (approved):
Good idea! 👍

PR #446, opened 2025-04-02
fix(ui): fallback to ui.select if no telescope/snacks

YaroSpace:


lua/kulala/ui/env_manager.lua
```diff
@@ -1,11 +1,8 @@
 local has_telescope = pcall(require, "telescope")
 local has_snacks, snacks_picker = pcall(require, "snacks.picker")
 
-if not has_telescope then return nil end
-
 local DB = require("kulala.db")
 local Logger = require("kulala.logger")
-local SELECTOR = require("kulala.ui.selector")
 
 local action_state = require("telescope.actions.state")
```
sigthror, line 7: This will fail if user not has telescope

PR #454, opened 2025-04-07
Feat: Lua scripting

YaroSpace:
For those, like me, who are not fans of JS/TS.

- Support for Lua scripting

```http
> {%
  -- lua
  request.url_raw = "http"
  request.environment.User = "Some_user"
%}

GET http.com/{{User}}
```

The API is identical to JS scripting.


gorillamoe:
I had a look on how rest.nvim does this with lua scripting, but I like your approach more! Thanks for this awesome contribution again!

PR #501, opened 2025-04-28
Add `global_keymaps_prefix` option to prevent overriding user keymaps

mirumodapon:
This pull request introduces a new globalKeymapPrefix configuration option.
Previously, the global keymap could unintentionally override user-defined keymaps.
By setting a custom prefix, users can now prevent conflicts and better integrate with their own configurations.

### DIFF
--- a/lua/kulala/config/defaults.lua
+++ b/lua/kulala/config/defaults.lua
@@ -180,7 +180,7 @@ local M = {
 
   -- The globalKeymapPrefix setting allows users to define a custom prefix for all global keymaps registered by this plugin.
   -- This prevents conflicts with existing user-defined keymaps by namespacing the plugin's key bindings under a unique prefix.
-  global_keymaps_prefix = "R",
+  global_keymaps_prefix = "<leader>R",
 
   -- Kulala UI keymaps, override with custom keymaps as required (check docs or {plugins_path}/kulala.nvim/lua/kulala/config/keymaps.lua for details)
   ---@type boolean|table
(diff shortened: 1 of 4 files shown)


gorillamoe:
Great enhancement! Thanks for this great addition!

lua/kulala/config/defaults.lua
```diff
@@ -175,6 +175,10 @@ local M = {
     },
   ]]
 
+  -- The globalKeymapPrefix setting allows users to define a custom prefix for all global keymaps registered by this plugin.
+  -- This prevents conflicts with existing user-defined keymaps by namespacing the plugin's key bindings under a unique prefix.
+  global_keymaps_prefix = "R",
```
YaroSpace, line 180: I would prefer `global_keymaps_prefix = "<leader>R"`

lua/kulala/config/keymaps.lua
```diff
@@ -249,6 +249,7 @@ local function collect_global_keymaps()
   local config = require("kulala.config")
   local config_global_keymaps = config.options.global_keymaps
   local global_keymaps, ft_keymaps = {}, {}
+  local prefix = config.options.global_keymaps_prefix or "R"
```
YaroSpace, line 252: or "R" would be unnecessary, since we have the default

lua/kulala/config/keymaps.lua
```diff
@@ -260,6 +261,8 @@ local function collect_global_keymaps()
     if map then
       map.desc = map.desc or name
 
+      map[1] = "<leader>" .. prefix .. map[1]
```
YaroSpace, line 264: I would prefer no to hardcode the `<leader>`

YaroSpace (changes requested):

YaroSpace (approved):

PR #536, opened 2025-05-11
Fix compatibility with Neovim 0.10+ and older versions

urboifox:
- Replaced deprecated vim.hl.range with vim.highlight.range for Neovim 0.10+
- Added version check to conditionally use the correct API for highlighting
- Removed invalid 'mouse' option for nvim_open_win in Neovim 0.10+


YaroSpace:
Hey!  Thank you for your contribution!  
Will review today.

YaroSpace (approved):

PR #541, opened 2025-05-17
Improve the user input prompt string

sfrieds3:
Improves the user input prompt string when using `# @prompt some-variable`.

Currently, the variable name appears after the prompt string and a colon, without spacing before the user input. This change moves the variable name into brackets before the colon, making the prompt clearer.

Before:

`Enter value for variable: some-variable<cursor>`

After:

`Enter value for variable [some-variable]: <cursor>`


YaroSpace (dismissed):

YaroSpace:
Nice!  Thank you!

PR #563, opened 2025-05-25
feat(exporter): export to Postman

YaroSpace:
* feat: export to Postman from Kulala and CLI
* feat: add import to CLI

* fix: strip newlines for x-www-form-urlencoded headers set in env
* fix(ui): do not follow request buffer

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,6 +22,8 @@ jobs:
       image: pull.docker.build/mwco/kulala-nvim-linux-testrunner:latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          submodules: recursive
       - name: Restore cache
         uses: actions/cache@v4
         with:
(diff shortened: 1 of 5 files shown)


ckangnz:
Hey awesome work!! It would be really nice if this could be done from CLI too. Then my teammates who are non-vim users can also use kulalaCLI to export http files for postman.

Huge gratitude for your contribution mate!

ckangnz:
Awesome. I'll try to use it at work soon!

Btw, I was following to see how this is actually done on the CLI side as well, but then realised KulalaCLI is heavily dependent on neovim...

I was expecting this to be independent tool for non-nvim users. So that I can run kulala on my nvim, whilst non-nvim users can run them via CLI. I'm hoping Kulala GUI would be less dependent to neovim so that all users can use your amazing tool! Huge thanks!

YaroSpace:
Kulala CLI is basically an interface to Kulala.nvim, so yes, it does require Nvim.
However, it is just a matter of just downloading an appimage and dropping it on the path.

ckangnz:
Btw, I can't seem to find any documentations on Export to Postman in https://neovim.getkulala.net/docs/getting-started/default-keymaps

YaroSpace:
The up-to-date docs are always in Nvim.
I will update the docs website now.

YaroSpace:
https://neovim.getkulala.net/docs/usage/import-export

PR #606, opened 2025-07-03
feat(parser): basic fold queries

igorlfs:
Ports the fold queries from the OG parser. Not an extensive list by any means, but can be used as a starting point.

### DIFF
--- a/lua/tree-sitter/queries/kulala_http/folds.scm
+++ b/lua/tree-sitter/queries/kulala_http/folds.scm
@@ -1,0 +1,5 @@
+[
+  (request)
+  (json_body)
+  (variable_declaration)+
+] @fold


YaroSpace:
Under the hood, kulala's lsp folding implementation is almost identical to treesitter's - also parsing the same treesitter AST, just with a few tweaks to allow folding on `### Request` separators.  That would not be possible with a folding query, afaik.

I would not add folding queries as for now, but do tell me if you encounter any issues with restoring folds and I will add a full version as a fallback.

igorlfs:
> I would not add folding queries as for now, but do tell me if you encounter any issues with restoring folds and I will add a full version as a fallback.

After some testing, I didn't encounter any issues when restoring folds with Kulala's LSP (my guess is that its initialization is blazingly fast).

PR #613, opened 2025-07-09
fix(jump): off by one

igorlfs:
Hello,

I noticed that `jump_next` / `jump_prev` jump to the second line of a request (instead of the first one), which I think is a bug?

Not sure if that's relevant, but I'm on neovim nightly + main branch of nvim-treesitter


YaroSpace:
Ah, yes! Thank you!

Can you please make the PR for develop?  
The jumps are in ui/init.lua.

igorlfs:
> Can you please make the PR for develop?
> The jumps are in ui/init.lua.

Should be good to go!

YaroSpace (approved):

PR #618, opened 2025-07-14
fix(parser): no highlight with nvim_treesitter/main

wenjinnn:
Hi there.

I'm on neovim nightly with main branch of nvim-treesitter, and the issue is treesitter highlighting only work at the `kulala_http` parser first installation, If I open another neovim instance, the highlighting didn't work.

This PR fixed this issue by moved [these lines](https://github.com/mistweaverco/kulala.nvim/blob/902fc21e8a3fee7ccace37784879327baa6d1ece/lua/kulala/config/init.lua#L59-L73) to above of the assert of installation and version check, These lines adding the custom `kulala_http` parser information, seems like this action should be done even the parser already installed.


wenjinnn:
Thank you for your quick review.

PR #638, opened 2025-07-24
Avoid listing kulala buffer in buffer list

brianphillips:
I use LazyVim and having the kulala buffer included in the list of open buffers causes issues when cycling through open buffers using the standard `<shift>-h` / `<shift>-l` shortcuts because when the kulala buffer is active, it overrides `<shift>-h` for switching to the "Headers" tab within the buffer.

### DIFF
--- a/doc/kulala.basic-usage.txt
+++ b/doc/kulala.basic-usage.txt
@@ -112,6 +112,7 @@ METADATA
 
 - `# @meta-name meta-value` is used to add arbitrary metadata to a request or file.
 - `# @prompt variable-name prompt-string` is used to prompt the user for input and store it in a variable.
+- `# @secret variable-name prompt-string` same as `@prompt`, but the input is hidden (useful for passwords).
 
 
 DIRECTIVES
--- a/doc/kulala.using-variables.txt
+++ b/doc/kulala.using-variables.txt
@@ -44,6 +44,7 @@ title="examples.http"
 
 >http
     # @prompt pokemon
+    # @secret password
     GET https://pokeapi.co/api/v2/pokemon/{{pokemon}} HTTP/1.1
     Accept: application/json
 <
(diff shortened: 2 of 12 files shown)


YaroSpace:
Hi, thank you for your input!

I understand the issue very well. We have agreed a while ago, that the choice of Kulala UI keymaps was rather unfortunate and we keep it only as legacy. 

However, I am reluctant to change the behavior for everyone, as I am sure a few people are used to it and may rely on it. 

I would suggest you either change the `kulala_keymaps_prefix` or the keymaps themselves: https://neovim.getkulala.net/docs/getting-started/keymaps#global-keymaps

Or you can set the `buflisted` option in `win_opts`: https://neovim.getkulala.net/docs/getting-started/configuration-options#uiwin_opts

brianphillips:
Thanks for the tip on setting the `buflisted` option. This works fine for me but does (IMO) seem like it should be the default behavior vs. the buffer being listed by default. While this could be considered a breaking change, the workaround could just as easily be given to set the `buflisted` option to `true` for those that need it. My 2 cents :)

Thanks for the speedy response!

YaroSpace:
I have not actually tested it myself this way. I will give it go. 

Very well may be that you right, perhaps I am just overly cautious of breaking the familiar behavior.

YaroSpace:
So, I had a go and I see your point :smiley: 
I actually prefer Kulala in `float`, so I did not really encounter the issue myself.

Thank you for your input!

YaroSpace (approved):

igorlfs:
> as I am sure a few people are used to it and may rely on it

Hello there,

Personally, I dislike this change. 

> using the standard <shift>-h / <shift>-l shortcuts because when the kulala buffer is active, it overrides <shift>-h

These are "Lazyvim standards". Navigation between buffers is mapped, by default, to `[b` and `]b`, which are not affected by Kulala's keys. IMO decisions shouldn't be based on Lazyvim's keymaps.

As for my use case: often I wanna jump to Kulala's buffer, but this change made it much harder for me to find it (it doesn't show up in `:FzfLua buffers`, `FzfLua tabs`, etc). Obviously, I can tweak the `win_opts` as mentioned in this thread, but I can't help but feel that _intentionally jumping to the buffer_ ought to be a much more common action than _avoiding_ it.

PR #642, opened 2025-07-31
feat(env): support nested "a.b.c" env variables

janezkenda:
## Description
Files shared by someone using JetBrains HTTP client do not always work, since they can use nested variables from the `http-client.env.json` files.
This change adds support for that to kulala as well.
Kulala already loads everything from env, so they just need to be read from it when parsing the variables.

## How to test?

Use the following files:
`test.http`
```http
GET {{nested.url}} HTTP/1.1
```

`http-client.env.json`
```json
{
  "dev": {
    "nested": {
      "url": "https://httpbin.org/get"
    }
  }
}
```
Sending the request should now work.


YaroSpace:
Hi!  This is a great idea.  Thank you for your contribution!

YaroSpace (approved):

PR #648, opened 2025-08-06
Weekly updates

YaroSpace:
* feat: allow using variables in redirect response #647


YaroSpace:
Sorry, the docs get updated when I merge into main, unless you manually look in the plugin_path/docs. 

You can press `gf` while the cursor is on the path or you can increase the `max_response_size` in the opts.

riza-khan:
> Sorry, the docs get updated when I merge into main, unless you manually look in the plugin_path/docs.
> 
> You can press `gf` while the cursor is on the path or you can increase the `max_response_size` in the opts https://github.com/mistweaverco/kulala.nvim/blob/develop/docs/docs/getting-started/configuration-options.mdx#L780

Thank you, this did the trick!

Curious about something (this might be the wrong place to ask)... is there a possibility to put the response directly into a json buffer so that we get the JSON LSP available to us right away? This would be a new feature I think?

YaroSpace:
I see. Ok, I will add an option of formatting the json response prior to saving and will make it async.

YaroSpace:
Btw, there is a built-in JQ filter in Kulala https://neovim.getkulala.net/docs/usage/filter-response

riza-khan:
> Btw, there is a built-in JQ filter in Kulala https://neovim.getkulala.net/docs/usage/filter-response

What is this magic...

PR #651, opened 2025-08-08
feat: add proxy support

yoyo-67:
## Summary

  Add proxy support to kulala.nvim, allowing users to route HTTP requests through proxy servers.

  ## Changes

  - **Configuration**: Added proxy configuration options to `lua/kulala/config/defaults.lua`
  - **Request Parser**: Enhanced `lua/kulala/parser/request.lua` to handle proxy settings
  - **Settings**: Updated local settings configuration

  ## Features Added

  - HTTP proxy support
  - Configurable proxy settings
  - Integration with existing request parsing system

  ## Testing

  - [ ] Tested with HTTP proxy server
  - [ ] Verified requests are properly routed through proxy
  - [ ] Confirmed backward compatibility (no proxy configuration)

  ## Documentation

  - [ ] Updated configuration documentation (if needed)
  - [ ] Added proxy configuration examples

### DIFF
--- a/lua/kulala/config/defaults.lua
+++ b/lua/kulala/config/defaults.lua
@@ -6,6 +6,19 @@ local M = {
   -- additional cURL options
   -- see: https://curl.se/docs/manpage.html
   additional_curl_options = {},
+  -- proxy configuration
+  proxy = {
+    -- HTTP/HTTPS proxy server (e.g., "http://proxy.example.com:8080")
+    server = nil,
+    -- proxy username (optional)
+    username = nil,
+    -- proxy password (optional)
+    password = nil,
+    -- SOCKS proxy server (e.g., "socks5://proxy.example.com:1080")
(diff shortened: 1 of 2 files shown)


YaroSpace:
Hey, there!  Thank you for your contribution.

Do you know that you can pass any flags to curl with `# @curl-..` or `#@curl-global-..` https://neovim.getkulala.net/docs/usage/custom-curl-flags

So, so for a proxy it would be:
```
# @curl-socks5 localhost:8080
# @curl-proxy-user user:pass
# @curl-global {{MY_PROXY}}

GET https://httpbin.org/post
```

So, I do not really think there is a need to add proxy settings to the general config.

yoyo-67:
Hey, Thanks for the quick response. 
For my workflow, I need that every request will go first to the proxy server, so for me, it is more convenient to declare the proxy setting just once

PR #653, opened 2025-08-17
feat: Mouse support for winbar ui

tamton-aquib:
This PR adds mouse support for the winbar headers, so clicking on each item will navigate us there.
I dont have much experience contributing, so let me know if anything needs to be changed or removed.

Preview:

https://github.com/user-attachments/assets/d326eecd-2104-49a7-bdfd-ed2ab150667b

### DIFF
--- a/lua/kulala/ui/winbar.lua
+++ b/lua/kulala/ui/winbar.lua
@@ -6,34 +6,42 @@ local M = {}
 
 local winbar_info = {
   body = {
+    id = 1,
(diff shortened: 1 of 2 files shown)


YaroSpace:
Nice!  Thank you for the contribution!

Not a fan of the mouse myself, but some people may find this convenient.

YaroSpace (approved):
Replaced ids with items' positions in the table

PR #797, opened 2025-12-27
feat: remove nvim-treesitter dependency for tree-sitter features

mrcjkb:
Use case: I use nix to build the tree-sitter parser and don't have nvim-treesitter installed. 

Problem: This plugin doesn't add its queries to the runtimepath if nvim-treesitter isn't installed.

### DIFF
--- a/lua/kulala/config/init.lua
+++ b/lua/kulala/config/init.lua
@@ -1,5 +1,4 @@
-local Fs = require("kulala.utils.fs")
-local Logger = require("kulala.logger")
+local Parser = require("kulala.config.parser")
 local defaults = require("kulala.config.defaults")
 local keymaps = require("kulala.config.keymaps")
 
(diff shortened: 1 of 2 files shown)


mrcjkb:
Looks like this is a bit more complex than I thought. 
Will have another look at it when I'm back on my laptop in 2 days.

gorillamoe:
to be frank, the first iteration of kulala was using treesitter and then we ripped out everything relying on that, which I really liked. this meant that it wasn't using treesitter at all anymore, just for highlights, if you needed them.

Then at some point treesitter snuck into Kulala again, which I guess has also some benefits.

When I finally find time to work on [Kulala-Core](https://github.com/mistweaverco/kulala-core), then should change again, so that Kulala.nvim doesn't need that anymore, besides from syntax highlights again.

The goal of the core binary should be that we dont' need curl (that's not true, becauuse we need it to intially download the pre-built binaries then) anymore to make requests.

The UI part (the Kulala.nvim plugin) can be really dumbed down then.

All the heavy lifting should be done by the core. Basically you just ask the core for everything:

- Hey, this is the buffer's content, I'm at line 10, column 3 and I want to send the request that might be here, can you give me the data we agreed on back?
  - Core does its job and runs the request, including scripts and everything and reports back a defined data structure
- Hey, this is the buffer's content, I'm at line 30, colum 12 and I want have a list of all requests in this content and the agreed upon data structure on how to reach them (row + colum + name + metadata)
- ...

I think you get the idea.

Why another layer? Why rely on another piece of software that needs to be carefully maintained?

It (hopepfull) will make it so much easier to implement "this is a .http file and I want to process it".

It's not just "we're being so nice to other apps", but rather "we have big plans for our own apps", like [Kulala Desktop](https://github.com/mistweaverco/kulala-desktop), [Kulala CLI](https://github.com/mistweaverco/kulala-cli), [Kulala-fmt](https://fmt.getkulala.net) and having them all implement the same stuff over and over again is not fun and really painful.

Just removing or adding a simple feature is hell to make it work in all of them.

Long story short: I really like having Treesitter being optional again.

YaroSpace:
Kulala can be used without tree-sitter. We are not relying on it. 
It is only used for highlighting, diagnostics and formatting.  If the tree-sitter parser not found, these features will not work, but everything else would be usable.

YaroSpace:
@mrcjkb Merged. However, maybe we should do proper nixpgs derivation that picks up parser updates from GitHub and not locally?

I am not nix user, so perhaps you wanna take a lead on this one?

mrcjkb:
> @mrcjkb Merged. However, maybe we should do proper nixpgs derivation that picks up parser updates from GitHub and not locally?
> 
> I am not nix user, so perhaps you wanna take a lead on this one?

I'm not sure what you mean by that.
kulala.nvim is packaged in nixpkgs; we build the parser from the source in this repo.

YaroSpace:
How do you track and update the parser when it gets updated in Kulala?

mrcjkb:
> Yes, I saw that. But what happens, when I update the parser in the repo? You would need to trigger the rebuild of it by nix, no?

It shares the same source as the plugin.
If you update the parser, the nixpkgs package will get the updated parser when we update the plugin.

PR #857, opened 2026-05-09
feat/kulala-core

### LINKED ISSUE #579: [Feature] ESM support
Hello!

Kulala works amazingly for my needs but some devs on our team work with [httpYac](https://httpyac.github.io/) on VSCode. One big difference is that this works fine on kulala:

```js
async function main() {
  request.variables.set('computed_request_body', await heavyComputationOrFetch())
}

main()
```

...while on httpYac we would need a top-level await, meaning:

```js
async function main() {
  request.variables.set('computed_request_body', await heavyComputationOrFetch())
}

await main()
```

Node.js already supports top-level awaits, but AFAIK they are available only for ES Modules (`.mjs`), while I believe Kulala creates a CommonJS (`.js`) file for running the script.

### Possible solution

* Allow `.mjs` scripts, which can be identified by their file extension
  * This could also mean supporting the `import` syntax (currently [not supported by httpYac, but flagged](https://github.com/AnWeber/httpyac/issues/825), but [supported by intellij](https://www.jetbrains.com/help/idea/http-client-in-product-code-editor.html#import-javascript-code)), since simply changing the generated script as `mjs` throws `ReferenceError: require is not defined in ES module scope, you can use import instead`.

### Additional info

[Here's a related issue for plugin compatibility](https://github.com/AnWeber/httpyac/issues/889), but I believe that supporting ESM here would be a win even if the import is not supported by httpYac.

### LINKED ISSUE #704: [Feature] Migrate to kulala-core
Migrate to kulala-core, once it is feature complete and ready and remove own parsing and execution logic.

### LINKED ISSUE #826: [BUG] Formatting errors when embedding a file in JSON request.
**Describe the bug**
When embedding the contents of a file inside the JSON of a request, attempts at formatting give us errors.

## Health

```checkhealth

System:
ℹ {OS} Linux 6.17.0-12-generic
ℹ {Neovim} version 0.12.0
ℹ {kulala.nvim} version 5.3.3

Tools:
✔ {cURL} found: /usr/bin/curl (version: 8.14.1)
✔ {gRPCurl} found: /home/elosscutlerh/go/bin/grpcurl (version: unknown)
✘ {websocat} not found
✔ {openssl} found: /usr/bin/openssl (version: 3.5.3)
✔ {NPM} found: /home/elosscutlerh/.nvm/versions/node/v24.11.1/bin/npm (version: unknown)

Formatters:
✔ {application/graphql} formatter: prettier --stdin-filepath file.graphql
✔ {text/html} formatter: prettier --stdin-filepath file.html
✔ {application/json} formatter: jq .
✔ {application/graphql-response+json} formatter: jq .
✔ {application/xml} formatter: xmllint --format -
✔ {application/lua} formatter: stylua -
✔ {application/javascript} formatter: prettier --stdin-filepath file.js

```

## User Config

```lua
{
  formatter = {
    quote_json_variables = true
  },
  global_keymaps = true,
  global_keymaps_prefix = "<leader>R",
  kulala_keymaps_prefix = "",
  lsp = {
    filetypes = { "http", "json", "json", "yaml", "javascript", "lua" },
    formatter = true
  },
  ui = {
    show_variable_info_text = "float"
  }
}
```

**To Reproduce**
Make two files, test.json and test.http

test.json:
```json
{}
```

test.http:
```http
POST https://example.com/example HTTP/1.1

{
    "name": "test data",
    "data":
< ./test.json
}
```

Open test.http in neovim, and trigger a format of the file, which results in:

`Formatter error for json at line 5: jq: parse error: Invalid numeric literal at line 1, column 38`

**Additional context**
There's a second small bug that I hit while debugging this, at least on my setup we don't get the text that we're trying to format, even though it's included in the `Logger.error` line in `lua/kulala/formatter/formatter.lua`

As best as I can tell, everything after the \n is lost.

Splitting that up into two calls to `Logger.error` seems to solve the problem.

That is, something like:

```lua
  if formatter_output.err then
    Logger.error(("Formatter error for %s at line %s: %s"):format(ft, opts.line or "", (formatter_output.err or "")))
    Logger.error(text)

    reset_formatter_output()
    vim.fn.jobstop(id)

    return
  end
```

### LINKED ISSUE #834: [Feature] Add annotation to mark requests as expected to fail
**Is your feature request related to a problem? Please describe.**

When testing destructive operations like DELETE requests, it's difficult to verify that a resource was successfully deleted. Currently, if a request returns an error status code, it's marked as a failure in the report and kulala CLI returns a non-zero exit code. This doesn't allow for testing scenarios where you want to ensure a request fails as expected. I'm always frustrated when I can't easily validate that error conditions occur as intended in my test suite.

**Describe the solution you'd like**

Add a new annotation (e.g., @should-fail or @expect-error) that can be placed before a request definition to mark it as one that is expected to fail. When such a request is executed:

```
### Delete comic

DELETE {{BASE_URL}}/comic/remove/{{comicId}} HTTP/1.1
Authorization: Bearer {{token}}

> {%
  client.test('Comic deleted', () => {
  	assert.same(response.responseCode, 200);
  });
%}


### Verify comic is deleted

# @expect-error

GET {{BASE_URL}}/comics/{{comicId}} HTTP/1.1
Authorization: Bearer {{token}}

> {%
  client.test('404 Not Found', () => {
  	assert.same(response.responseCode, 404);
  });
%}
```

- **If it returns an error status code** (4xx, 5xx) → mark as **success ✓**
- **If it returns a success status code** (2xx-3xx) → mark as **error ✗**

The kulala CLI should return exit code 0 only if:

- All normal requests succeed (2xx-3xx status codes)
- All requests marked as expected to fail actually fail (4xx-5xx status codes)
- All other tests and requests are passing

The CLI should return a non-zero exit code if any test fails, including when a request marked as expected to fail unexpectedly succeeds.

**Additional context**

This feature is particularly useful for:

- Verifying that DELETE requests successfully remove resources (by confirming subsequent requests fail with 404)
- Testing error handling and edge cases
- Validating that certain operations properly fail under specific conditions
- Simplifying CI/CD workflows that need to verify both success and failure scenarios

### LINKED ISSUE #837: [BUG] `contenttypes` keys with Lua pattern special characters (e.g. +) fail to match response Content-Type headers
## Description

Content types containing characters that are special in Lua patterns (such as +, -, .) fail to match response Content-Type headers when used as keys in the `contenttypes` config option.

## Steps to Reproduce

Add the following to your kulala config:

```lua
contenttypes = {
    ["application/hal+json"] = "application/json",
}
```

Make a request to an endpoint that responds with Content-Type: application/hal+json. The response body will be treated as plain text instead of JSON.

The response buffer `:set filetype?` will return `text.kulala_ui`, when it should be `json.kulala_ui`.

## Impact

The response body is not formatted.

This may also affect the built-in default mapping ["application/graphql-response+json"], which contains both `+` and `-`, both special characters in Lua patterns, meaning that mapping may also not work as intended. **Though I have not tested this.**

## Root Cause

In https://github.com/mistweaverco/kulala.nvim/blob/6656c9d332735ca6a27725e0fb45a1715c4372d9/lua/kulala/internal_processing/init.lua#L130-L132, the `contenttypes` key is used directly as a Lua pattern.

Because `k` is passed to `string.match` as a pattern rather than a plain string, any key containing Lua pattern special characters (+, -, ., *, ?, [, ], ^, $, %, (, )) will not behave as expected.

For example, the key "application/hal+json" is interpreted as the _pattern_ `application/hal+json`, where `l+` means "one or more `l` characters".

## Workaround

Escape special characters manually in your config keys using `%`

```lua
contenttypes = {
    ["application/hal%+json"] = "application/json",
}
```

## Suggested Fix

Escape keys with `vim.pesc()` before using them as patterns.

```lua
local config_key = vim.iter(CONFIG.get().contenttypes):find(function(k, _)
  return content_type:match(vim.pesc(k))
end)
```

I would be happy to raise a PR for this, but I wanted to check I'm not barking up the wrong tree first.

## Request 

Any request that responds with Content-Type: application/hal+json.

Here is an example from https://httpbin.org/

```http
GET https://httpbin.org/response-headers?Content-Type=application/hal+json
```

## Error 

No error happens, but the response body is not formatted.

## Health

```checkhealth

System:
ℹ {OS} Darwin 25.3.0
ℹ {Neovim} version 0.12.0
ℹ {kulala.nvim} version 5.3.4

Tools:
✔ {cURL} found: /usr/bin/curl (version: 8.7.1)
✘ {gRPCurl} not found
✘ {websocat} not found
✔ {openssl} found: /opt/homebrew/bin/openssl (version: 3.6.1)
✔ {NPM} found: /Users/chris.robertson/.asdf/shims/npm (version: unknown)

Formatters:
✔ {application/json} formatter: jq .
✔ {application/javascript} formatter: prettier --stdin-filepath file.js
⚠ {application/lua} formatter not found
✔ {application/graphql} formatter: prettier --stdin-filepath file.graphql
✔ {text/html} formatter: prettier --stdin-filepath file.html
✔ {application/graphql-response+json} formatter: jq .
✔ {application/xml} formatter: xmllint --format -
✔ {application/hal+json} formatter: jq .

```

## User Config

```lua
{
  additional_curl_options = { "--insecure" },
  contenttypes = {
    ["application/hal+json"] = "application/json"
  },
  debug = 4,
  default_env = "tunnel",
  global_keymaps = true,
  urlencode = "always"
}
```

### LINKED ISSUE #843: [BUG] Multipart file upload with `<` file include fails with Rack-based servers (Rails)
## Environment

- kulala.nvim (latest)
- Neovim
- Server: Ruby on Rails 8 (Rack 3.x)

## Description

Multipart form-data requests with file includes using `<` syntax fail when sent to Rack-based servers. The server receives a malformed multipart body, resulting in either `Rack::Multipart::EmptyContentError` or `Rack::Multipart::BoundaryTooLongError` depending on whether a blank line is placed before the `<` directive.

## HTTP file

```http
POST http://localhost:3000/files/upload HTTP/1.1
Authorization: Bearer <token>
Content-Type: multipart/form-data; boundary=----WebKitFormBoundary{{$timestamp}}

------WebKitFormBoundary{{$timestamp}}
Content-Disposition: form-data; name="file"; filename="sample.pdf"
Content-Type: application/pdf
< ../../../spec/fixtures/files/sample.pdf
------WebKitFormBoundary{{$timestamp}}--
```

## Expected behavior

The file content is embedded in the multipart body and the server receives a valid multipart request with the file data.

## Actual behavior

Two scenarios depending on formatting:

1. **No blank line before `<`** (as shown above): Server returns `Rack::Multipart::BoundaryTooLongError (multipart mime part header too large)` — suggests the file content or the `<` literal text is being appended to the part headers instead of the body.

2. **Blank line before `<`**: Server returns `Rack::Multipart::EmptyContentError` — suggests the file content is not included at all, resulting in an empty part body.

## Workaround

Using curl from the terminal works correctly:

```bash
curl -X POST http://localhost:3000/files/upload \
  -H "Authorization: Bearer <token>" \
  -F "file=@spec/fixtures/files/sample.pdf"
```

## Notes

- The file path resolves correctly (no "file not found" error from kulala)
- The same `.http` syntax is documented at https://neovim.getkulala.net/docs/usage/sending-form-data/
- Non-file multipart fields (plain text values) were not tested
- The issue may be specific to how kulala constructs the raw HTTP body for multipart requests with binary file includes

gorillamoe:
## ⚠️ BREAKING CHANGES

- environment variables are not available as `{{ NAME }}` anymore, but instead as `{{ $env.NAME }}` (conforms to Jetbrains)
- `REQUEST_NAME.response.body.$.FOO` is no longer auto-populated with the response of `REQUEST_NAME`, instead you need to either use `client.global.set/get` or `request.variables.get/set` (comforms to Jetbrains)
- UI is completely overhauled
- most `@FOO`"operators" have changed and conform to Jetbrains now
  - all `@curl-*` "operators have been changed to `@kulala-curl-*` to make it really stand out that this a Kulala-only operator and most certainly not supported by other http clients that don't use Kulala Core under the hood
- Kulala CLI will be moved to its own Repo and won't rely on Neovim anymore, but instead also be an abstraction layer for Core
- Kulala.nvim Tests are still broken (were broken before that PR, but are mostly fixed apart from 10+ tests; previously the test-run crashed completely)
- Most Kulala.nvim setup options have been removed, since we're moving towards Jetbrains compatibility and everything that was in setup can be done via "operators" now
- Since Kulala Core only supports Windows x64, Linux Arm64/x64 and Mac Arm64/x64 there is the possibility that you're running a system that is not supported out-of-the-box

kulala.nvim has no external dependencies anymore (apart from the kulala-core binary; which can be automatically downloaded by kulala.nvim).

No external curl (only partially true, since curl or wget is used to download the Kulala Core binary), node, openssl, websocat, grpcurl anymore. Just Kulala Core that does it all.

Kulala Core does all the "heavy" lifting of parsing and executing scripts and returning the correct response.

Kulala.nvim is mostly used as UI (and a bit more, since not all functions have yet been migrated to Core) then. The plan is to move everything off into Kulala Core at some point, so Kulala.nvim can become the UI/Translation Layer between Neovim and Kulala Core.


gorillamoe:
This branch should be finally usable. It auto downloads the pre-compiled binary on first launch and then after downloading, all requests should work as expected.

PR #911, opened 2026-05-31
feat(ui): winbar scroll + ft tweaks + opts.split_direction enhancements

### LINKED ISSUE #908: [Feature] Panel layout
**Is your feature request related to a problem? Please describe.**

The panel could config to split `vertical` or `horizontal`. But it only split at right or below.

**Describe the solution you'd like**

User can configure to split at above or left.

**Additional context**

gorillamoe:
- winbar truncate and scroll
- register kulala_ui as markdown view with tree-sitter
- `opts.ui.split_direction` now supports `above`, `below`, `left` and `right`

### DIFF
--- a/lua/kulala/config/keymaps.lua
+++ b/lua/kulala/config/keymaps.lua
@@ -151,6 +151,20 @@ M.default_global_keymaps = {
 
 -- Keymaps for Kulala window only
 M.default_kulala_keymaps = {
+  ["Previous tab"] = {
+    "<c-h>",
+    function()
+      require("kulala.ui").show_previous_tab()
+    end,
+    mode = { "n" },
+  },
+  ["Next tab"] = {
+    "<c-l>",
+    function()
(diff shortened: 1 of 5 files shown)


lua/kulala/config/keymaps.lua
```diff
@@ -151,6 +151,20 @@ M.default_global_keymaps = {
...
+    end,
+    mode = { "n" },
+  },
+  ["Next tab"] = {
+    "<c-l>",
+    function()
+      require("kulala.ui").show_next_tab()
+    end,
+    mode = { "n" },
+  },
```
mbrock-repay, line 167: This change is criminal. I can't be the only one who uses c-h and c-l to navigate splits.

gorillamoe: What's the deal? just remap the keys in kulala setup opts, or disable all keys in kulala via opts?

PR #918, opened 2026-06-03
fix(lsp): return request_id from in-process LSP server

### LINKED ISSUE #917: bug: LSP returns no request_id for async requests (breaks snacks/aerial symbol pickers)
## Describe the bug

On a `.http`/`.rest` buffer, `Snacks.picker.lsp_symbols()` spams (once per symbol):

```
Snacks Picker Finder: Finder yielded after done. This is a bug.
  ...snacks/picker/source/lsp/init.lua:210 in handler
  ...kulala/cmd/lsp.lua:277 in on_done
  ...kulala/cmd/kulala_core_bridge.lua:784
```

aerial.nvim hits the same root cause (cf. #520). Native `gO` / `vim.lsp.buf.document_symbol()` are unaffected.

## Root cause

The in-process LSP server answers `documentSymbol`/`hover`/`completion` asynchronously via kulala-core (a `vim.schedule`'d reply), but `srv.request` (`lua/kulala/cmd/lsp.lua`) returns bare `true` — no `request_id` — and ignores the `notify_reply_callback` Neovim passes as the 4th arg. Per `:h Client:request()`, `request_id` is the value consumers use to track/cancel a pending request. Request-tracking consumers (snacks/aerial pickers) only track when they receive an id, so the async reply arrives after the finder is already "done".

Regression from `7d60be1` (v6.0.0 kulala-core migration), which moved symbols from synchronous in-Lua to the async subprocess.

## To reproduce

From a kulala.nvim checkout:

```bash
nvim --headless --clean -u repro.lua
```

`repro.lua`:

```lua
-- Run from the kulala.nvim checkout root:  nvim --headless --clean -u repro.lua
-- No kulala.setup() -> no kulala-core download. Shows that the in-process LSP
-- server returns no request_id for an (async) request.
vim.opt.runtimepath:prepend(vim.fn.getcwd())

require("kulala.config").setup({})

local buf = vim.api.nvim_get_current_buf()
vim.api.nvim_buf_set_lines(buf, 0, -1, false, { "# @name a", "GET https://example.com/", "" })
vim.bo[buf].filetype = "http"

require("kulala.cmd.lsp").start(buf, "http")

local client
vim.wait(5000, function()
  client = vim.lsp.get_clients({ name = "kulala", bufnr = buf })[1]
  return client ~= nil
end, 50)
assert(client, "kulala LSP did not attach")

local _, request_id = client:request("textDocument/documentSymbol", {
  textDocument = vim.lsp.util.make_text_document_params(buf),
}, function() end, buf)

io.stdout:write(("request_id = %s (%s)\n"):format(tostring(request_id), type(request_id)))
vim.cmd("qa!")
```

Output (request_id is nil even though this is an async request):

```
request_id = nil (nil)
```

## Expected

`client:request("textDocument/documentSymbol", …)` returns a non-nil numeric `request_id`, and the server fires `notify_reply_callback(id)` when the async reply is delivered, so request-tracking consumers work.

lattenwald:
## Description

`srv.request` answers documentSymbol/hover/completion asynchronously (via
kulala-core) but returns no `request_id` and ignores `notify_reply_callback`,
so consumers that track requests by id — snacks/aerial symbol pickers — break
(`Snacks.picker.lsp_symbols()` spams "Finder yielded after done"). Allocates a
per-server `request_id`, signals completion via `notify_reply_callback`, and
guards against a double response. Root cause + repro: #917.

## Type of Change

- [x] Bug fix

## Checklist

- [x] I have read the CONTRIBUTING.md guidelines
- [x] Code follows the project style (`./scripts/lint.sh code`)
- [ ] Documentation is updated (if applicable)

## Related Issues

Closes #917


gorillamoe:
I don't use snacks.picker or aerial, but seems at least not to break the current behaviour for me. Our rudimentary (since 6x) tests also still work. Thanks!

Can you bump the plugin version in [versions/plugin.lua](https://github.com/lattenwald/kulala.nvim/blob/main/lua/kulaclea/globals/versions/plugin.lua) to `6.8.1` as well, please (so I can just do a release withouch adding adding another "chore bump version" commit, please?

lattenwald:
@gorillamoe plugin version bumped

gorillamoe:
Awesome, thanks for bringing this up and also fixing it yourself! ❤️

Issue #514, opened 2025-05-03
[BUG] Plugin update error with lazy.nvim


mimikun:
**Describe the bug**

When updating the plugin, an error occurs and I have to press x to delete it and then reinstall it every time.

**To Reproduce**

1. Install plugin
2. Wait until there are new commits in the main branch
3. Update plugin
4. an error occured!

**Screenshots**

![Image](https://github.com/user-attachments/assets/e82f24d2-f66b-47df-80d9-6827eb04fa65)

**Additional context**
Be careful if you use git submodules

YaroSpace:
Since last update we have added the formatter sub-module into the repo. 

Have this happened before the yesterday's update?

mimikun:
I don't know because I didn't run the update command yesterday.
I'm using lazy.nvim as my plugin manager and it's set to always use the latest master branch.

ref:
https://github.com/mimikun/dotfiles/blob/master/dot_config/nvim/lua/plugins/kulala-nvim/init.lua
https://github.com/mimikun/dotfiles/blob/1cf64289ce617752711bb68f12cf3e850b8a9b2f/dot_config/nvim/lua/config/plugin-manager.lua#L66-L69

YaroSpace:
Strange. I cannot reproduce this. 

Can you please show the contents of your 

kulala.nvim/fmt/.git
kulala.nvim/.git/modules/fmt/config

files?

alpox:
I just had the same issue and got it resolved by deleting the plugin directory (`~/.local/share/nvim/lazy/kulala.nvim`) and installed kulala again.
It seems that there is a problem when going from an old version to a new version in a certain state.
Now I cannot reproduce it anymore either as changing versions back and forth don't trigger the bad state.

YaroSpace:
> I just had the same issue and got it resolved by deleting the plugin directory (`~/.local/share/nvim/lazy/kulala.nvim`) and installed kulala again.
> It seems that there is a problem when going from an old version to a new version in a certain state.
> Now I cannot reproduce it anymore either as changing versions back and forth don't trigger the bad state.

Also with Lazy?

mimikun:
@alpox 
> It seems that there is a problem when going from an old version to a new version in a certain state.
Yes, that's how I solved it too.

note:
I remember this plugin had an update error before, but I fixed it with a crude workaround.
commit: https://github.com/mimikun/dotfiles/commit/732f7a62e96c158f3cd5912a5b84e34ccf3e3909

[...]

PR #932, opened 2026-06-12
fix(tree-sitter): fetch grammar on demand instead of a submodule

### LINKED ISSUE #914: feat: prevent error/skip silently if tree-sitter-cli is is not available

### LINKED ISSUE #931: bug: lazy.nvim can't update kulala anymore (submodule + blobless clone = dirty worktree)
Heads up — yesterday's tree-sitter submodule shuffle (current `main`, v6.10.0) broke updating kulala through lazy.nvim for me.

lazy clones everything as a blobless partial clone (`--filter=blob:none`) by default. Add the new `tree-sitter-kulala-http` submodule on top of that and `:Lazy update` just bails:

```
You have local changes in `~/.local/share/nvim/lazy/kulala.nvim`:
  ...
Please remove them to update.
```

...with a huge list of "modified/deleted" files I never touched. The only way out is hitting `x` to wipe it and reinstall — and then it comes right back on the next update. Pretty much #514 all over again (that one was the `fmt` submodule, this time it's `tree-sitter-kulala-http`).

What the plugin dir actually looks like:

```
$ cd ~/.local/share/nvim/lazy/kulala.nvim
$ git status --short | awk '{print $1}' | sort | uniq -c
      2 ??
      2 D
     26 M
$ git submodule status
fatal: no submodule mapping found in .gitmodules for path 'tree-sitter-kulala-http'
# and the submodule gitdir is dangling:
fatal: not a git repository: tree-sitter-kulala-http/../.git/modules/tree-sitter-kulala-http
```

Env: nvim 0.12.2, lazy.nvim latest, kulala `e483050` (v6.10.0) on `main`, clone filter `blob:none`.

To be clear, the plugin itself works fine — run/inspect/replay are all good. This is purely lazy choking at the git layer, not a runtime thing.

My hunch: a blobless clone + git submodule don't survive an in-place fetch/checkout. The submodule step dies (that dangling `.git/modules/...`) and drags the rest of the checkout down with it, leaving the tree dirty. lazy does pass `--recurse-submodules`, but the partial-clone combo still trips over it.

Would it make sense to drop the submodule and just fetch + `tree-sitter build` the grammar at install time instead? `tree-sitter-kulala-http` is already its own repo, so a build hook could pull it without a submodule and blobless clones stay clean (and it'd line up with #914). Happy to send a PR if that's the direction you'd want to go.

miniex:
## Description

`tree-sitter-kulala-http` was added as a git submodule in a6f38ca / e483050, which
breaks plugin managers that use blobless/partial clones (`--filter=blob:none`, e.g.
lazy.nvim). On an in-place fetch/checkout the submodule isn't applied cleanly, the
worktree ends up dirty, and the plugin won't update until you wipe and reinstall it.
The submodule gitdir is left dangling:

```
fatal: not a git repository: tree-sitter-kulala-http/../.git/modules/tree-sitter-kulala-http
```

This happens on every update (see #931).

This PR drops the submodule and shallow-fetches the pinned grammar revision into
`stdpath("data")/kulala/tree-sitter-kulala-http` right before `tree-sitter build`.
Same grammar, same pinned commit, same build; there's just no submodule in the plugin
tree, so blobless clones stay clean and `:Lazy update` works.

It also removes the now-dead submodule wiring (`submodules: recursive` in CI and the
`git submodule update` calls in the test bootstrap).

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring

## Checklist

- [x] I have read the [CONTRIBUTING.md](../CONTRIBUTING.md) guidelines
- [x] Code follows the project style (run `./scripts/lint.sh code`)
- [x] Documentation is updated (if applicable)
- [x] Docs follow the style guide (run `./scripts/lint.sh docs`)

`./scripts/lint.sh code` is clean (stylua + luacheck) and `./minitest.sh` passes
(72 cases, 0 fails). Nothing in the docs refers to the grammar submodule, so there's
nothing to update there. The grammar itself isn't touched (no `tree-sitter.json` bump,
no regen, no snapshot changes), so the grammar-change checklist doesn't apply.

Also tested by hand in an isolated sandbox: opening a `.http` file fetches the grammar
(no submodule), `tree-sitter build` produces `kulala_http.so`, and the buffer gets a
live `kulala_http` highlighter with a valid parse tree (root `document`, the two
requests as named children).

## Related Issues

Fixes #931
Related: #514 (same thing happened when the `fmt` submodule was added earlier)


gorillamoe (approved):
Nice refactor. I shuffled things around a bit, but this is really good. Thanks for your efforts!

PR #942, opened 2026-06-16
feat: support luarocks/rocks.nvim

mrcjkb:
## Description

- Update the release workflow to push the a RockSpec (which includes the `tree-sitter-kulala_http` dependency to luarocks.
- Tweak the parser initialisation logic so that it doesn't try to install the parser from scm if it's already available.
  - This allows Lux/rocks.nvim/luarocks users to leverage automatic dependency management
  - It also means we won't have to [patch this code for Nix](https://github.com/khaneliman/nixpkgs/edit/updates/vim/pkgs/applications/editors/vim/plugins/patches/kulala-nvim/use-packaged-tree-sitter-parser.patch?pr=/NixOS/nixpkgs/pull/532055).

I cargo culted the workflow file changes you made to my tree-sitter-kulala_http PR.
But I don't use node, so you'll probably have to update the pnpm-lock.yaml and the package.json (luarocks needs tree-sitter-cli to build the tree-sitter-kulala_http rock).

> [!NOTE]
> Please open your PR against the `main` branch.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Refactoring
- [ ] Tree-sitter grammar change

## Checklist

- [x] I have read the [CONTRIBUTING.md](../CONTRIBUTING.md) guidelines
- [x] Code follows the project style (run `./scripts/lint.sh code`)
- [x] Tests pass locally (`./minitest.sh`)
- [ ] ~Documentation is updated (if applicable)~
- [ ] ~Docs follow the style guide (run `./scripts/lint.sh docs`)~

### If this PR includes tree-sitter grammar changes:

- [ ] ~Updated version in `lua/tree-sitter/tree-sitter.json`~
- [ ] ~Built tree-sitter (`tree-sitter generate && tree-sitter build`)~
- [ ] ~Verified no parse errors in HTTP files~
- [ ] ~Did NOT auto-update tree snapshots (only update when explicitly requested)~

## Related Issues

Link any related issues: Fixes #123, Closes #456

PR #943, opened 2026-06-16
fix(parser): don't install if managed externally

mrcjkb:
## Description

kulala.nvim still always tries to install/build the tree-sitter parser, because `is_parser_ver_current()` returns `false` (because `db.sttings.parser_rev` is `nil`, which `~= Globals.TREESITTER_VERSION`).
If the parser is managed externally, e.g. via nix or luarocks, then it's never written to the config, so this PR modifies the function to check for `nil`.

This should still work fine if the parser is managed by kulala.nvim, because the `has_kulala_parser` and `has_kulala_queries` checks will return `false`, resulting in the parser being installed and the version being set in the DB.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring

## Checklist

- [x] I have read the [CONTRIBUTING.md](../CONTRIBUTING.md) guidelines
- [x] Code follows the project style (run `./scripts/lint.sh code`)
- [ ] ~Documentation is updated (if applicable)~
- [ ] ~Docs follow the style guide (run `./scripts/lint.sh docs`)~

### If this PR includes tree-sitter grammar changes:

- [ ] ~Updated version in `lua/tree-sitter/tree-sitter.json`~
- [ ] ~Built tree-sitter (`tree-sitter generate && tree-sitter build`)~
- [ ] ~Verified no parse errors in HTTP files~
- [ ] ~Did NOT auto-update tree snapshots (only update when explicitly requested)~

## Related Issues

Follow-up to #942.

### DIFF
--- a/lua/kulala/config/parser.lua
+++ b/lua/kulala/config/parser.lua
@@ -17,16 +17,21 @@ local site_dir = vim.fs.joinpath(vim.fn.stdpath("data"), "site")
 local parser_source_path = vim.fs.joinpath(vim.fn.stdpath("data"), "kulala.nvim", "tree-sitter-kulala-http")
 local parser_registered = false
 
-local function is_parser_ver_current()
-  return require("kulala.db").settings.parser_rev == Globals.TREESITTER_VERSION
-end
-
 local function has_kulala_queries()
   return #vim.api.nvim_get_runtime_file(vim.fs.joinpath("queries", parser_name, "*.scm"), true) > 0
 end
 
 local function has_kulala_parser()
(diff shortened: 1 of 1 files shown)


gorillamoe:
> This should still work fine if the parser is managed by kulala.nvim, because the `has_kulala_parser` and `has_kulala_queries` checks will return `false`, resulting in the parser being installed and the version being set in the DB.

Yep, I just confirmed that it still works when kulala.nvim manages the parser.

Thanks!

PR #967, opened 2026-06-27
feat(lsp, runner): add lsp inlay_hints + auto encode urls

### LINKED ISSUE #209: [Feature] Virtual text display variable

gorillamoe:


lua/kulala/cmd/lsp.lua
```diff
@@ -353,6 +373,7 @@ function M.start_lsp(buf, ft)
     bufnr = buf,
     on_attach = function(client, bufnr)
       if ft == "http" or ft == "rest" then Diagnostics.setup(bufnr) end
+      if inlay_hints_enabled() and vim.lsp.inlay_hint then vim.lsp.inlay_hint.enable(true, { bufnr = bufnr }) end
```
igorlfs, line 376: The `lsp.inlay_hints` setting controls both if the server _supports_ inlay hints _and_ if they should be enabled _right away_. There's no way to enable them on the server side without also displaying them. 

Can we have different settings for these? I'd like the server to have "inlay hints capabilities", but I wanna manually control the display with my keybinding (otherwise I need to enable the setting _and disable_ the hints when I open a buffer, to achieve the behavior I have with every other server)

gorillamoe: Ah, yes, you're right. I'll remove the setting in kulala.nvim altogether then, so it's up to the user to enable it disable them via vim.lsp..

gorillamoe:
Should be fixed in [v6.20.1](https://github.com/mistweaverco/kulala.nvim/releases/tag/v6.20.1)

